PR #19, opened 2019-05-31
add extras for service and test related dependencies

imatiach-msft:
This PR adds extras for service and test related dependencies.
We want to separate out the core algorithm related dependencies from those that are used for the dashboard and testing, which shouldn't be required in all scenarios.
The user can include extras by specifying them when installing the package, eg:

pip install interpret[dashboard,testing]

The default install command would not include them:

pip install interpret


interpret-ml:
Hi @imatiach-msft - good to see you again, and thanks for the first PR!

On our end we'll get the CI system public facing first, and will get back to you.

imatiach-msft:
An alternative might be to create two packages, one with the core algorithms that other packages can depend on and another with the visualization/dashboard/tests for users to try out the package

interpret-ml:
Agreed, there seems to be an alternative for separating out core algorithms with visualization. 
For now, we'll merge your code into a separate branch, and do some adjustments so that we can make testing optional. Thanks for the PR!

PR #36, opened 2019-07-17
Eedeleon/add blackbox extension

eedeleon:
Add an extension hook for blackbox explainers.

Downstream libraries can use the extension hook to add classes under the interpret.ext.blackbox namespace if validation is passed

Add an example class registration.

TODO
- decide on proper validation for blackbox explainers
- decide on a more useful example blackbox explainer

### DIFF
--- a/src/python/interpret/ext/blackbox/__init__.py
+++ b/src/python/interpret/ext/blackbox/__init__.py
@@ -1,0 +1,19 @@
+import logging
+import sys
+
+from interpret.ext.extension_utils import load_class_extensions
+
+module_logger = logging.getLogger(__name__)
+
+BLACKBOX_EXTENSION_KEY = "interpret_ext_blackbox"
+
+
+def _is_valid_blackbox_explainer(proposed_blackbox_explainer):
+    return True
+
+
+# How to get the current module
+# https://stackoverflow.com/questions/1676835
+current_module = sys.modules[__name__]
+
+load_class_extensions(current_module, BLACKBOX_EXTENSION_KEY, _is_valid_blackbox_explainer)
--- a/src/python/interpret/ext/blackbox/__init__.py
+++ b/src/python/interpret/ext/blackbox/__init__.py
@@ -1,3 +1,6 @@
+# Copyright (c) 2019 Microsoft Corporation
+# Distributed under the MIT software license
+
 import logging
 import sys
 
--- a/src/python/interpret/ext/blackbox/__init__.py
+++ b/src/python/interpret/ext/blackbox/__init__.py
@@ -12,7 +12,10 @@ BLACKBOX_EXTENSION_KEY = "interpret_ext_blackbox"
 
 
 def _is_valid_blackbox_explainer(proposed_blackbox_explainer):
-    return True
+    for explanation_type in ["local", "global", "perf", "data"]:
+        if hasattr(proposed_blackbox_explainer, "explain_" + explanation_type):
+            return True
+    return False
 
 
 # How to get the current module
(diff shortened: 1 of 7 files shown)


src/python/interpret/ext/extension_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    Used to validate class names before registration.
+
+    :param proposed_class_name: The string to name the class
+    :type proposed_class_name: str
+    """
+    # regex for class name came from
+    # https://stackoverflow.com/questions/10120295
+    # TODO: Support other languages
+    match = re.match(r"[a-zA-Z_][a-zA-Z_0-9]+", proposed_class_name)
```
wamartin-aml: Any need to impose a max length for any part of the naming of this stuff? I guess it only makes life worse for the person who's writing the extension, so probably no need

src/python/interpret/ext/extension_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    if match is None or match.group(0) != proposed_class_name:
+        raise ValueError("Invalid class name {}. Class names must start with a "
+                         " letter or underscore. And can continue with letters, "
```
wamartin-aml: nit: a little cleaner as "Class names must start with a letter or an underscore and can continue with letters, numbers, and underscores."

src/python/interpret/ext/blackbox/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+
+from interpret.ext.extension_utils import load_class_extensions
+
+module_logger = logging.getLogger(__name__)
+
+BLACKBOX_EXTENSION_KEY = "interpret_ext_blackbox"
+
+
+def _is_valid_blackbox_explainer(proposed_blackbox_explainer):
+    return True
```
wamartin-aml: The idea here is we build this out as we figure out exactly what our parameters for blackbox explainers are?

src/python/interpret/ext/blackbox/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+# How to get the current module
+# https://stackoverflow.com/questions/1676835
+current_module = sys.modules[__name__]
+
+load_class_extensions(current_module, BLACKBOX_EXTENSION_KEY, _is_valid_blackbox_explainer)
```
wamartin-aml: Worth thinking about moving the validation function to be called separately? Just for simplification

src/python/interpret/ext/blackbox/__init__.py
eedeleon: Maybe, we noticed an issue with validation and required parameters.

For now validation is converging to has explain_local or explain_global defined.

We are still figuring out if it is possible to instantiate the class with default args and require a _get_validation_init if the default instantiate would not work.

src/python/interpret/ext/blackbox/__init__.py
eedeleon: I liked having one function call for error handling to be simpler but breaking this up could avoid function passing which people don't always like.
I can look into breaking this up into a few function calls:

to_register = get_proposed_class_extensions
valid_classes = [class for class in to_register if _is_valid]
register_class_extensions(valid_classes)

src/python/interpret/ext/extension_utils.py
eedeleon: Currently no, and it will fail on creation if python has an internal limit

interpret-ml:
Looks pretty good so far. Agree on the TODO, we can do that our end if it's easier.
There's a major change coming to the directory structure so what we'll do is merge this into a `extensions` branch. Once the refactor is done we'll figure out the remaining TODOs and then merge into `develop` for next release.

PR #41, opened 2019-07-31
Add multi-class post-processing

zhangxz1123:
multi-class post-processing added to
interpret/src/python/interpret/glassbox/ebm/multiclass_postprocess.py.

Currently omitting the center step, since we haven't discussed the input & output format for the centering. It is assumed that the centering will take a string list containing 'categorical' or 'numerical'. But what is the output format? I assume we then have a list 2D array updated_feature_graphs as well as a 1D array intercept?

### DIFF
--- a/src/python/interpret/glassbox/ebm/multiclass_postprocess.py
+++ b/src/python/interpret/glassbox/ebm/multiclass_postprocess.py
@@ -1,13 +1,15 @@
 import copy
 
-def multiclass_postprocess(X_binned, feature_graphs, binned_predict_proba):
+def multiclass_postprocess(X_binned, feature_graphs, binned_predict_proba, feature_types):
     '''
     X_binned: Training dataset, pre-binned. Contains integer values, 0+. Each value is a unique bin.
     feature_graphs: List of 2d numpy arrays. List is size d for d features. Each numpy array is of size b for b bins in the feature. Each bin has k elements for k classes.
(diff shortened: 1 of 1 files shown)


interpret-ml:
Hi @zhangxz1123, 

Thanks for the PR! Just wanted to address your comments -- with the inclusion of the centering step, we'll be able to merge this in. Everything else looks good!

>  It is assumed that the centering will take a string list containing 'categorical' or 'numerical'.

That's correct, we'll eventually pass you this as an input parameter. You can assume the new function signature will look like: 

```python
def multiclass_postprocess(X_binned, feature_graphs, binned_predict_proba, feature_types):
```

"feature_types" will be a list of strings containing either "categorical" or "numeric" for each feature.

--------

>  I assume we then have a list 2D array updated_feature_graphs as well as a 1D array intercept

Yes, this sounds right. You can return these as a lightweight dictionary 

```python
return { 'feature_graphs' : updated_feature_graphs, 'intercept' : intercept}
```
where updated_feature_graphs is a list of 2d numpy arrays, and intercept is a 1d array.

zhangxz1123:
Done adding centering. Please review. Thx.

interpret-ml:
Hi Xuezhou, 

Thanks again for the PR! Everything looks good to us. We'll be adding in an experimental version of multiclass EBM to the next release, with your postprocessing included. Really appreciate all your hard work!

PR #53, opened 2019-08-22
Add interaction purification utility

blengerich:
In additive models, effect sizes can be freely moved between the matrix representing interaction effects and the main effects.  To make sure we are consistent when interpreting interaction effects, it would be great to present models in an identifiable form. We can accomplish this goal by "mass-moving" any interaction effects with non-zero row means or non-zero column means into the corresponding main effect plots.

This PR includes a utility file purify.py and associated tests to perform "mass-moving" on numpy arrays. It is not currently integrated with the EBM API.

### DIFF
--- a/python/interpret/utils/__init__.py
+++ b/python/interpret/utils/__init__.py
@@ -3,3 +3,4 @@
 
 from .all import *  # noqa: F401,F403
 from .distributed import *  # noqa: F401,F403
+from .purify import *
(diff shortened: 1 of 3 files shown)


blengerich:
Added the link to our paper about purifying interaction effects: [https://arxiv.org/abs/1911.04974](https://arxiv.org/abs/1911.04974). 

This PR can be merged.

interpret-ml:
Hi @blengerich - thanks for the PR!

This has been merged in for our next release. The code has been moved into `interpret.glassbox.ebm.research` module where you can access all the purification calls.

PR #59, opened 2019-09-13
adding model-specific glassbox explainer extension namespace

imatiach-msft:
adding model-specific glassbox explainer extension namespace

Note: this was for whitebox models due to reason below, but I changed it to glassbox because whitebox (or graybox) is already being done on a separate branch

There needs to be a separate extension mechanism for model-specific explainers such as SHAP DeepExplainer, LinearExplainer and TreeExplainer, which don't fall under the blackbox extension

### DIFF
--- a/python/interpret/ext/glassbox/test/test_example_glassbox_explainer_ext.py
+++ b/python/interpret/ext/glassbox/test/test_example_glassbox_explainer_ext.py
@@ -1,0 +1,21 @@
+# Copyright (c) 2019 Microsoft Corporation
+# Distributed under the MIT software license
+
+from ....glassbox import LinearRegression
+from ....api.base import ExplainerMixin
+from .. import _is_valid_glassbox_explainer
+
+
+def test_invalid_glassbox_explainer():
+    # NOTE: Available method claims local exists, but there is not respective method.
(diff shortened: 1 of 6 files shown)


python/interpret/ext/glassbox/test/test_example_glassbox_explainer_ext.py
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) 2019 Microsoft Corporation
+# Distributed under the MIT software license
+
+from ....glassbox import LinearRegression
+from ....api.base import ExplainerMixin
+from .. import _is_valid_glassbox_explainer
+
+
+def test_invalid_glassbox_explainer():
```
wamartin-aml: I get that this is a bit more difficult, but can we at least test that one of the current glassbox explainers in the interpret package will pass our checks?

imatiach-msft: sure, I can make a separate PR with that test, would like to get this in first

interpret-ml:
Hi @imatiach-msft / @wamartin-aml,

Looks good. There's a refactor for the extension system in the latest branch which will collide with this. We're going to merge this in and make it work with the rest of the system. Thanks for your help!

PR #60, opened 2019-09-16
fix namespace collisions with interpret package

imatiach-msft:
fix namespace collisions with interpret package

I really only need the change in the top-level  python/interpret/\_\_init\_\_.py, but I made it consistent across all init files.
The issue I was running into is that with another package, when using the same interpret.community namespace, it could not find my package and gave me errors like:

```
    from interpret.community.tabular_explainer import TabularExplainer
E   ModuleNotFoundError: No module named 'interpret.community'
```

relevant stack overflow:

https://stackoverflow.com/questions/2699287/what-is-path-useful-for

### DIFF
--- a/python/interpret/__init__.py
+++ b/python/interpret/__init__.py
@@ -22,3 +22,5 @@ logging.getLogger(__name__).addHandler(NullHandler())
 
 # Set name of package
 name = "interpret"
+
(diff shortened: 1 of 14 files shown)


imatiach-msft:
closing since we decided not to have packages under the same interpret namespace

PR #64, opened 2019-09-23
Float sens prediction vals

Ashton-Sidhu:
Hi,

I am proposing to cast the prediction array being passed to the morris analyze function to float as morris requires the array to be float values.

The reason for this is for the `.predict` function of sklearn classifiers return a list of integers causing the global explainer morris technique to throw an error.

Whether or not it makes more sense to use `.predict` vs. `.predict_proba`, all other explainer analysis methods pass when using `.predict`, so this change will make it so that all explainer analysis methods will work with `.predict_proba` and `.predict` when explaining sklearn classifiers.

EDIT: Also removed 2 print statements to clean up output when using the shap local explainer.

Ashton

### DIFF
--- a/python/interpret-core/interpret/blackbox/sensitivity.py
+++ b/python/interpret-core/interpret/blackbox/sensitivity.py
@@ -1,14 +1,13 @@
 # Copyright (c) 2019 Microsoft Corporation
 # Distributed under the MIT software license
 
-from abc import ABC, abstractmethod
-
-import numpy as np
-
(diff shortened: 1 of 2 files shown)


interpret-ml:
Hi @Ashton-Sidhu,

Thanks for the PR, looks great!
Ignore the build error, it's a flaky integration test.

Agree with the comments, it should work for predict and predict_proba where possible. This will be going in the next release. Nice work!

PR #66, opened 2019-09-24
Choose main attributes

ncherrier:
Add an argument to choose which attributes will be used as the main attributes instead of systematically using all of them. This could be useful especially when the user has already performed feature selection, and/or is also specifying interaction pairs.

### DIFF
--- a/python/interpret-core/interpret/glassbox/ebm/ebm.py
+++ b/python/interpret-core/interpret/glassbox/ebm/ebm.py
@@ -331,6 +331,7 @@ class BaseCoreEBM(BaseEstimator):
         col_types=None,
         col_n_bins=None,
         # Core
+        main_attr="all",
         interactions=0,
         holdout_split=0.15,
         data_n_episodes=2000,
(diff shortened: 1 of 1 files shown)


interpret-ml:
Hi @ncherrier, 

Thanks for the pull request! We originally imagined that users would subset their numpy arrays/dataframes before passing it into EBM, which would then cause EBM to only train on the subsetted features. 

> especially when the user has already performed feature selection, and/or is also specifying interaction pairs

That being said, do you mind elaborating more on this? We were hoping to understand your use case a bit better.

Thanks!
-InterpretML Team

ncherrier:
Hi @interpret-ml ,
I am actually performing feature selection ahead of EBM on both main features and pairs of features. In this way, some features may appear in pairs but not in the main attributes. In this particular setting, I don't want these features to be used as main attributes.

interpret-ml:
Hi @ncherrier,

This sounds logical. Thanks for elaborating on this.

PR looks good, we'll merge, do a few more checks and then have it in next release!

PR #71, opened 2019-10-01
Add checks to ensure model has been fitted before running explain global/local

chhetri22:
Called the check_is_fitted() function from the sklearn library to ensure that user receives a helpful error message when they try to get the local or global explanation without fitting the model


interpret-ml:
Hi @chhetri22,

Looks good. This will be in the next release. Thanks for the PR!

PR #91, opened 2019-11-06
Fix bug to correctly set VisualizeProvider ip address and port

mczhu:
The ip address and port set through interactive.set_show_addr()
are not passed to AutoVisualizeProvider (and DashProvider). As a result,
interactive.show() does not render at the user defined address.

Also add unit tests to assert that the ip and port are set correctly.


interpret-ml:
Hi @mczhu - apologies for the delay.

This looks really solid: good changes, good tests. LGTM.
Will be in this upcoming release (should be this week).

PR #110, opened 2020-02-12
(nit) doc fix for missing param in ebm

imatiach-msft:
(nit) doc fix for missing param in ebm


interpret-ml:
Hi Ilya - thanks for the PR!

Looks good, merging for next release.

PR #111, opened 2020-02-12
added initial scipy sparse conversion to python layer and test

imatiach-msft:
Added initial scipy sparse conversion to python layer and test

This adds trivial scipy sparse support by converting the data to dense in unify data.  Adding real sparse support to be done in the future.


interpret-ml:
Hi Ilya - thanks for the PR!

Good clean-up of our duplicated code btw - merging for next release.

PR #120, opened 2020-04-10
Port interpret-community Permutation Feature Importance to interpret

eedeleon:
# Add PermutationImportance Explainer class with support for global explanations.

.visualize is implemented with LinearExplainer's visualize function and a subset of the parameters.

## Worth noting
New API with required labels: I am passing data and labels in PermutationImportance constructor, similar to EBM.fit.

Explain local seems to receive X and y in some cases, and no param in others. I could not find an Explain Global or Constructor that requiredin X and y as params.

I have added a condition within test_explainers.py to special case PermutationImportance.


eedeleon:
Meeting review:

- default metric that can be used for either. Mean Squared Error
Metric lambda, like predict_fn  sklearn metric spec with string support for sklearn metrics.


explain_global(name) caches summary when fit or constructor was called.

python/interpret-core/interpret/blackbox/permutationimportance.py
```diff
@@ -0,0 +1,210 @@
...
+                         'fbeta_score',
+                         'precision_score',
+                         'recall_score'}
+
+
+class PermutationImportance(ExplainerMixin):
+    available_explanations = ["global"]
+    explainer_type = "blackbox"
+
+    def __init__(
```
wamartin-aml, line 50: Can we get documentation for functions?

python/interpret-core/interpret/blackbox/permutationimportance.py
```diff
@@ -0,0 +1,210 @@
...
+                              predict_function, true_labels, base_metric, global_importance_values):
+        # Create a copy of the original dataset
+        shuffled_dataset = np.array(dataset, copy=True)
+        # Shuffle one of the columns in place
+        shuffled_dataset[:, col_idx] = shuffled_dataset[random_indexes, col_idx]
+        idx = col_idx
+        self._add_metric(predict_function, shuffled_dataset, true_labels,
+                         base_metric, global_importance_values, idx)
+
+    def explain_global(self, name=None):
```
wamartin-aml, line 124: It's very weird to me that data goes in the constructor, effectively making this a one-time use object. Why is that the case?

python/interpret-core/interpret/blackbox/permutationimportance.py
eedeleon: Currently interpret does not have much constructor docs. I decided to drop docs instead of highlighting the distinction on this class.

I do agree a doc addition PR should be made, but I think it should add docs to groups of explainers instead of one off additions.

wamartin-aml: There also aren't any on functions though and I know they have those

python/interpret-core/interpret/blackbox/permutationimportance.py
eedeleon: This is the convention within the repo. I do think this should be looked into with a design review for validation with external repos. However, for this PR I believe keeping to convention is the more correct approach.


Long winded answer :
When compared to glassbox models, it is consistent, since fit is only called once. However, you are right, creating two global explanations for comparison requires two constructor calls and two explain_global calls which seems off.

My initial implementation had explain_global(X, y) 
# similar to ebm.fit(X, y) and *.explain_local(X, y)

In general I agree that a convention should be chosen. I prefer only constructor or only explain_global. While reviewing explainers for consistency in both repos I saw a few interpret-community explainers with both constructor data and explain_global data requirements in a subset of explainers. That really confused me.

My preference after the review tended toward explain_global(X, y, name=None), with no data passed to any constructor. It felt more in line with explain_local(X, y) and fit(X, y) APIs found within interpret. It also could help with creating an Explainer, serializing it, and then explaining global on datasets on different machines with different data.

Either way, I think keeping to the repo convention is the correct approach for this PR. And a github issue for reviewing this convention would be worthwhile, even to just gather supporting evidence for the convention chosen.

wamartin-aml: Sure, consistency for this PR makes sense. The constructor vs. explain_global data in interpret-community is train vs. test data, at least as far as I'm aware. We don't require an explicit `fit` call from the user, but we assume that they'll want training data to be consistent while test data will vary.

python/interpret-core/interpret/blackbox/permutationimportance.py
eedeleon: I found no docs in https://github.com/interpretml/interpret/blob/master/python/interpret-core/interpret/blackbox

Which docs are you referring to?

python/interpret-core/interpret/blackbox/permutationimportance.py
eedeleon: I see. So you can call explain global on different test sets with the init on training data. 

That makes some sense. Definitely worth bringing up in the discussion

python/interpret-core/interpret/blackbox/permutationimportance.py
wamartin-aml: Wow ok you're right - I could have sworn there used to be more in the form of this lonely one I found: https://github.com/interpretml/interpret/blob/master/python/interpret-core/interpret/glassbox/ebm/ebm.py#L118

interpret-ml:
LGTM

PR #146, opened 2020-07-08
fix environment detection for azureml notebook VM case

imatiach-msft:
fix environment detection for azureml notebook VM case

For AzureML notebook VM case, we just need to check if this file exists to detect that we are in this environment:
/mnt/azmnt/.nbvm


interpret-ml:
LGTM - thanks for the contribution!
Ignore the build failure it's an internal issue.

PR #192, opened 2021-01-08
Make multiple show example more clear

raethlein:
The change should make it more obvious that you can pass the explanations of multiple models to the show function, not the actual models.


interpret-ml:
Hi @raethlein,

Good point, merging in now. Thanks!

-InterpretML Team

PR #195, opened 2021-01-15
Fix rendering in Azure ML from 2nd run onwards

jruales:
Azure ML, like Colab, uses individual iframes per output so it needs init_js every run.


DouglasXiao (approved):

interpret-ml:
Hi @jruales, @DouglasXiao

Interesting, thanks for catching this! Will merge into next release.

-InterpretML Team

Issue #172, opened 2020-10-12
Query on show() method.


prateekiiest:
Hi,

For my case, after installing the package, the show Function doesnt seem to work and I get no output in the jupyter notebook.

I am currently running everything on my local notebook.

Kindly let me know if I am missing out something

interpret-ml:
Hi @prateekiiest --

Could you please run your code in debug mode and send us the log?  Also, which browser are you running this in?

```python3
from interpret.develop import debug_mode
debug_mode(log_filename='log.txt', log_level='INFO', native_debug=False)
```

-InterpretML team

lokijota:
I'm actually also having this issue, and it's pretty random, but mostly I don't get an output.

I'm running the notebooks in an Azure ML Compute Instance VM (an NC6), using either Jupyter/JupyterLab or the default Notebook experience in AML, and sometimes I get a display out, sometimes I don't. The cells always run without errors. I've tried different browsers (Edge Dev and Edge Beta), and while I don't get errors in the browser console, like I said -- sometimes it shows everything and fully functional, but 99% of the times it doesn't, the show() just doesn't work.

I ran the command to log errors, all I get is a single entry:

```
2020-12-10 21:54:11,130 | develop.py           159                 debug_mode() | {
  "interpret.__version__": "0.2.1",
  "interpret.status_show_server": {
    "app_runner_exists": false
  },
  "interpret.static_system_info": {
    "platform": "Linux-4.15.0-1100-azure-x86_64-with-debian-stretch-sid",
    "platform.architecture": [
      "64bit",
      ""
    ],
    "platform.machine": "x86_64",
    "platform.processor": "x86_64",
    "platform.python_version": "3.6.9",
    "platform.release": "4.15.0-1100-azure",
    "platform.system": "Linux",
    "platform.version": "#111~16.04.1-Ubuntu SMP Thu Nov 19 06:49:21 UTC 2020",
    "psutil.logical_cpu_count": 6,
    "psutil.physical_cpu_count": 6,
    "psutil.virtual_memory.total": "55.0GiB",
    "psutil.swap_memory.total": "0.0B"
  },
  "interpret.dynamic_system_info": {
    "psutil.virtual_memory": {
      "total": "55.0GiB",
      "available": "52.7GiB",
      "percent": 4.3,
      "used": "1.7GiB",
      "free": "51.2GiB",
      "active": "1.9GiB",
      "inactive": "1.1GiB",
      "buffers": "214.0MiB",
[...]

PR #210, opened 2021-03-14
Fix rendering for IPython/Jupyter based on env detection

prateekiiest:
- Provision for cases where both cloud and non-cloud env is detected.

Solves issue #172 

Screenshot (196)

Screenshot (195)


prateekiiest:
Hi @interpret-ml Team,

I hope this solves the long term issue https://github.com/interpretml/interpret/issues/172

Let me know if there's any improvement I can make here.

Cheers,
Prateek

prateekiiest:
my MS account doesn't have access to the interpret-internal build, so if the build issue can be checked and be let known here, that would be great.

prateekiiest:
Any updates on this @interpret-ml or should I close this?

python/interpret-core/interpret/provider/visualize.py
```diff
@@ -37,17 +37,35 @@ def _lazy_initialize(self):
...
             )
             self.provider = InlineProvider(detected_envs=detected_envs, js_url=JS_URL)
         elif "docker-dev-mode" in detected_envs:
             log.info("Operating in docker development mode.")
             self.provider = InlineProvider(detected_envs=detected_envs)
+
+        elif self.in_cloud_env == ENV_DETECTED.BOTH_CLOUD_AND_NON_CLOUD:
+            log.info("Detected both cloud and non cloud environment.")
+            val = input("Type 'C' if you want to choose Cloud environment or 'NC' for Non Cloud Environment :")
+            if val == 'C':
```
paulbkoch, line 55: Thanks @prateekiiest -- The only change I'd suggest is that instead of calling input(...) we choose the cloud option by default when both are present since input waits for a user response.

paulbkoch:
Thank you @prateekiiest for this fix!

prateekiiest:
Thanks @paulbkoch for the merge. Happy to contribute

PR #224, opened 2021-04-14
Adding predict_proba_and_explain to ExplainableBoostingClassifier

msplants:
This new method decreases runtime of a `predict` + `explain_local` by 4x by reducing the number of calls to calculate scores and returning explanations in a raw format (i.e. a numpy array).


interpret-ml:
Hi @msplants,

This looks great! Thanks for this PR, what we'll do is merge this into `develop`, and rename `predict_and_explain` to `predict_and_contrib` in order to prevent confusion with outputs from `explain_*`. This should be scheduled for next release.

Issue #239, opened 2021-05-20
Prevent users from feeding duplicate interaction terms


xiaohk:
Hello, I am training a `ExplainableBoostingRegressor` model. One of its parameters `interactions` can be a list of lists of feature indices. However, it seems users can include duplicate interaction terms.

For example, one can create a model with `interactions=[[0, 1], [0, 1], [1, 0]]`, and the trained model has three interaction terms with different `feature_importances_` and different `additive_terms_`. As these three terms all refer to the same interaction term, I guess we can handle this edge case by checking and removing duplicate terms in the `interactions` argument.

interpret-ml:
Nice catch -- thanks for reporting this @xiaohk! This would be a great fix for a first time contribution if you're interested in making a PR, but we can also address it in an upcoming release.

PR #240, opened 2021-05-21
Prevent users from feeding duplicate interaction terms (#239)

xiaohk:
Prevent users from feeding duplicate interaction terms (#239)


interpret-ml:
Hi @xiaohk,

This change looks good. Thanks for the PR!

PR #245, opened 2021-05-31
Replacing symbolic links

luisffranca:
When cloning the code on Windows the symlinks in symbolic/* are cloned as files, which leads to errors when running 'make install'. 

With this PR, instead of using symlinks the files are copied to symbolic/* during the build.


interpret-ml:
These look great! Merging to `develop` branch.

PR #265, opened 2021-07-26
Add conversion functions for GAM Changer

xiaohk:
Add functions to communicate with GAM Changer

- [x] Generate model data file
- [x] Generate sample data file
- [x] Modify Python EBM model based on `.gamchanger` file
  - Currently we only support load modified continuous variable (users can only change continuous variables in GAM Changer now)


paulbkoch:
Closing this PR since we ended up putting GAMChanger into its own separate repo under the InterpretML org:

https://github.com/interpretml/gam-changer
https://interpret.ml/gam-changer/

PR #313, opened 2022-01-01
Loosen ipython and ipykernel minimum version

eddy-geek:
... to remain compatible with Google Colab.
The only API used is from IPython `display` and `display_html`


paulbkoch:
Thank you @eddy-geek!

PR #335, opened 2022-04-29
Adding Composite terms importance

luisffranca:
This PR adds utility functions to compute composite terms importances and append them to Global Explanations.


interpret-ml:
Looks great.  Thanks @luisffranca!

PR #348, opened 2022-07-06
Adding EBM Feature Importances Notebook

luisffranca:


examples/python/notebooks/EBM Feature Importances.ipynb
```diff
@@ -0,0 +1,435 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from interpret import show\n",
+    "index = 0\n",
+    "show(ebm.explain_local(X[index:index+1], y[index:index+1]))"
```
msplants, line 138: Just curious: why are you selecting the first row this way? I think using `head` or `iloc` would be the more typical way of getting the first row:

```
show(ebm.explain_local(X.head(1), y.head(1)))
```

msplants (approved):
Overall, this looks great. Just a few nits really. Thanks!

examples/python/notebooks/EBM Feature Importances.ipynb
luisffranca: I'm doing this way because it was suggested that we provide code that could be used for any index, and not only the first one.

PR #371, opened 2022-10-11
Create offset.py

spyrosUofA:
Adding offset to EBM Classifier and Regressor (and other dependencies) 

Signed-off-by: Spyros <72393668+spyrosUofA@users.noreply.github.com>

### DIFF
--- a/python/interpret-core/interpret/glassbox/ebm/offset.py
+++ b/python/interpret-core/interpret/glassbox/ebm/offset.py
@@ -1,0 +1,1935 @@
+import interpret
+from interpret.glassbox.ebm.internal import Native
+from typing import DefaultDict
+from interpret.provider.visualize import PreserveProvider
+from interpret.utils import gen_perf_dicts
+from interpret.glassbox.ebm.utils import DPUtils #, EBMUtils
+from interpret.glassbox.ebm.postprocessing import multiclass_postprocess
+from interpret.utils import unify_data, autogen_schema, unify_vector
+from interpret.api.base import ExplainerMixin
+from interpret.api.templates import FeatureValueExplanation
+from interpret.provider.compute import JobLibProvider
+from interpret.utils import gen_name_from_class, gen_global_selector, gen_local_selector
+import ctypes as ct
+from multiprocessing.sharedctypes import RawArray
+import numpy as np
+from warnings import warn
+from sklearn.base import is_classifier
+from sklearn.utils.validation import check_is_fitted
+from sklearn.metrics import log_loss, mean_squared_error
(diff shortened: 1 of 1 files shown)


paulbkoch:
Hi @spyrosUofA -- Thanks for this PR!  I had a look over it and have a few comments:

I see the automated tests are failing on this line where offset.py is attempting to reference the EBMPreprocessor class https://github.com/spyrosUofA/interpret/blob/a81310bd21c8fc23efee28a4cede44e0780a82c8/python/interpret-core/interpret/glassbox/ebm/offset.py#L45

The EBMPreprocessor class had been moved to bin.py at this point in development:
https://github.com/spyrosUofA/interpret/blob/a81310bd21c8fc23efee28a4cede44e0780a82c8/python/interpret-core/interpret/glassbox/ebm/bin.py#L1509

And in the most recent code, it has been moved again to this location:
https://github.com/interpretml/interpret/blob/df8cf74ce26de4242a6660a6e3a179f8c610295b/python/interpret-core/interpret/utils/_binning.py#L1819


If my understanding is correct the "Off" parameter functions similarly to the "init_score" parameter in LightGBM ( https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.LGBMClassifier.html#lightgbm.LGBMClassifier ).  If that is correct, I would recommend using the LightGBM naming in order to make it easier to share calling code between the two and potentially leverage existing user knowledge of their parameters.

At a higher level, the current PR replicates a lot of code that exists in other locations within the codebase.  I understand why this was done since a lot has changed in the preprocessing section.  It would be better in terms of maintainability though to try and re-use as much of the other code as possible by passing init_score through the existing utility functions.  In the same vein I think init_score could, similarity to LightGBM, be a parameter to the fit function for all classes that derive from the EBMModel class instead of creating the new OffsetExplainableBoostingRegressor and OffsetExplainableBoostingClassifier classes.

One section of our code that I recommend reading over is the interaction detection code.  We have a similar requirement to use existing scores there, and do very similar processing:
https://github.com/interpretml/interpret/blob/df8cf74ce26de4242a6660a6e3a179f8c610295b/python/interpret-core/interpret/utils/_interaction.py#L54

spyrosUofA:
Hi @paulbkoch, thank you for your reply. Indeed, this offset code was built for v0.27. @tony-mtl was able to add the offset to v0.30 and addressed the above comments. Committing the new code now!

paulbkoch:
Closing this PR since your newer PR that adds init_score https://github.com/interpretml/interpret/pull/426 has been merged.

PR #391, opened 2022-12-22
SPOT: prototype selection 

bamdevm:
We contribute the SPOTgreedy code and a python notebook to show experiments with the code. Both the files are in the folder "interpret/python/interpret-core/interpret/utils/SPOT"


bamdevm:
There was signoff error on my part.

PR #392, opened 2022-12-22
SPOT greedy 

bamdevm:
prototype selection and the corresponding notebook.


bamdevm:
The files are in "interpret/python/interpret-core/interpret/utils/SPOT"

PR #425, opened 2023-04-19
Added Objective / Link Functions (Poisson Deviance Log Link, Gamma Deviance Log Link, MSE Log Link)

mtl-tony:
The addition of some added objective functions. This include Gamma and Poisson which have both been tested. Poisson is also included with an L2 loss as well. Poisson and Gamma have both been tested and benchmarked.

Also added the inv_link function that needs to be updated with new loss functions.


paulbkoch:
This is really great!  Thanks @mtl-tony!

PR #426, opened 2023-04-22
Adding init_score to predict, fit, ebm_decision_function etc..

mtl-tony:
Including init_score is a way to fit in which we have an initial score to boost on top of. This is currently tested for vectors. Not sure if it would work with other models. Also referred to as offset in GLM methodologies. Meant to allow fitting on top of predictions for other models as well.


paulbkoch:
Thanks @mtl-tony!  This is a great new feature for EBMs to have.

PR #436, opened 2023-05-15
add streamlit as a detected_envs

### LINKED ISSUE #423: Streamlit Provider
Hi **InterpretML teams**, 
Thank you very much for the amazing work. 

I’m trying to integrate Global and Local Explainer into my dashboard Streamlit. 
Unfortunately, function `show()` doesn’t have an option returning `HTML raw`, which can be work with **streamlit** using `streamlit.components.v1.html()`

Here is my resolution : 
```python
set_visualize_provider(InlineProvider(detected_envs = ['streamlit']))
```
In function `render()` of `interpret.visual.inline` : 
```python
elif "streamlit" in detected_envs:
    import streamlit.components.v1 as components
    components.html(HTML(init_js + body_js).data, height=1000, width=1000, scrolling=True)
 ```
image

It works perfectly for me. I don’t know if there’s a cleaner way to do this.
Also, I can make PR if you find it can be useful for other people.
Many thanks!
Dat LQ.

### PR DESCRIPTION
**Pull request by quocdat-le-insacvl**
This PR fixes #423. 
This is my first contribution to open-source. I hope to receive a review.

### CODE REVIEW

File: python/interpret-core/interpret/visual/_inline.py [hunk 1]
```diff
@@ -243,6 +244,8 @@ def render(explanation, id_str=None, default_key=-1, detected_envs=None, js_url=
         or "azuresynapse" in detected_envs
     ):
         display(HTML(init_js + body_js))
+    elif "streamlit" in detected_envs:
```
paulbkoch, line 247: My main question is whether we can avoid adding the streamlit dependency for all installations.  Does it still work if the import streamlit.components.v1 statement is put inside this elif "streamlit" block?  Presumably if "streamlit" is detected in the environment the streamlit dependency is installed?

quocdat-le-insacvl: Yes, you're right, it works fine.
I'm going to create a new PR and close this one because I didn't do it properly.

PR #438, opened 2023-05-16
add streamlit as environment Signed-off-by: Author Name quocdat-le-insacvl@gmail.com

### LINKED ISSUE #423: Streamlit Provider

quocdat-le-insacvl:
This PR fixes #423

### DIFF
--- a/python/interpret-core/interpret/visual/_inline.py
+++ b/python/interpret-core/interpret/visual/_inline.py
@@ -243,6 +243,9 @@ def render(explanation, id_str=None, default_key=-1, detected_envs=None, js_url=
         or "azuresynapse" in detected_envs
     ):
         display(HTML(init_js + body_js))
+    elif "streamlit" in detected_envs:
+        import streamlit.components.v1 as components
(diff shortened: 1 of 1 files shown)


paulbkoch:
Thanks @quocdat-le-insacvl!  It's great that we now have streamlit support.

PR #469, opened 2023-08-11
new methods

brandongreenwell-8451:
This PR addresses https://github.com/interpretml/interpret/issues/460 by adding two new methods to a fitted EBM object: 
* `reweight_terms()` - which allows callers to reweight the individual term contributions of the fit;
* `remove_terms()` - which allows callers to remove specific terms (and their associated components).

### DIFF
--- a/python/interpret-core/interpret/glassbox/_ebm/_ebm.py
+++ b/python/interpret-core/interpret/glassbox/_ebm/_ebm.py
@@ -2208,8 +2208,8 @@ class EBMModel(BaseEstimator):
                 floats whose i-th element corresponds to the i-th element of the
                 ``.term_*_`` attributes (e.g., ``.term_names_``).
             remove_nil_terms: Logical indicating whether or not to automatically
-                remove all terms that provide zero contribution to the fit 
-                (e.g., any term with a weight of zero).
+                remove all terms that are given a weight of zero; terms with a
+                weight of zero provide zero contribution to the fit.
 
         Returns: 
             Itself.
@@ -2262,8 +2262,8 @@ class EBMModel(BaseEstimator):
 
         # Delete "nil" terms (i.e., terms providing zero contribution to the fit)
         if remove_nil_terms:  # should automatically catch zero weight terms
-            # Grab indices of terms with zero contribution
-            term_list = np.where(self.term_importances() == 0)[0].tolist()
+            # Delete components that have a weight of zero
+            term_list = np.where(weights == 0)[0].tolist()
             return self.remove_terms(term_list)
         else:
             return self
--- a/python/interpret-core/interpret/glassbox/_ebm/_ebm.py
+++ b/python/interpret-core/interpret/glassbox/_ebm/_ebm.py
@@ -2146,10 +2146,10 @@ class EBMModel(BaseEstimator):
         return self
 
     def remove_terms(self, terms):
-        """Removes terms (and their associated components) from a fitted EBM. Note 
-        that this will change the structure (i.e., by removing the specified 
+        """Removes terms (and their associated components) from a fitted EBM. Note
+        that this will change the structure (i.e., by removing the specified
         indices) of the following components of ``self``: ``term_features_``,
-        ``term_names_``, ``term_scores_``, ``bagged_scores_``, 
+        ``term_names_``, ``term_scores_``, ``bagged_scores_``,
         ``standard_deviations_``, and ``bin_weights_``.
 
         Args:
@@ -2161,12 +2161,15 @@ class EBMModel(BaseEstimator):
         check_is_fitted(self, "has_fitted_")
 
         # If terms contains term names, convert them to indices
-        terms = [self.term_names_.index(term) if isinstance(term, str) 
-                else term for term in terms]
+        terms = [
+            self.term_names_.index(term) if isinstance(term, str) else term
+            for term in terms
+        ]
 
         def _remove_indices(x, idx):
             # Remove elements of a list based on provided index
             return [i for j, i in enumerate(x) if j not in idx]
+
         term_features = _remove_indices(self.term_features_, idx=terms)
         term_names = _remove_indices(self.term_names_, idx=terms)
         term_scores = _remove_indices(self.term_scores_, idx=terms)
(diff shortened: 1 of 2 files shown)


brandongreenwell-8451:
@paulbkoch it was just easier to start afresh and create a new PR (I'm also not the most Git/GH savvy of statisticians...). This version places the `remove_nil_terms` argument over to the `reweight_terms()` method and internally calls `remove_terms()` when set to `True`.

**Note:** This works great on my end, but not sure if the return statement on line 2267 is the most Pythonic way to handle the nested call to `remove_terms()`.

brandongreenwell-8451:
> Hi @brandongreenwell-8451 -- We definitely need some kind of remove_terms function that only operates on the terms given to it by the caller. The eventual ISLE function does not need this, but EBMs do as a general ability, so I figure it's best to have it as a primitive that is called by reweight.
>
>Another small nuance: I think the reweigh_terms function should only remove terms that have 0 in the given weight parameter, not check if any of the resulting term scores are all zero. If some of the terms had all zeros in their scores before the call to reweigh_terms, then the caller would need to do a lot of checks to know which terms to expect will be dropped, and that behavior would be surprising, I think. Safer to force the caller to be more explicit in this case and call remove_terms themselves in the special case that there are terms with all zeros and they want them removed. It's pretty unlikely we'd see a term with all zeros naturally anyways, except if the user previously edited the EBM (in which case whatever they called to do that could also have a remove_nil_terms parameter), or in the case of a feature with only 1 value. In the latter case LASSO would give it a weight of 0 anyways I think, so those terms would be dropped anyways in an ISLE implementation.

I think we're aligned on your first comment in this PR, right? Copy on the second and will send an update shortly!

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -2145,6 +2145,129 @@ def monotonize(self, term, increasing="auto", passthrough=0.0):
 
         return self
 
+    def remove_terms(self, term_list):
```
paulbkoch, line 2148: I'd like to call this "terms" instead of "term_list" and accept any enumerable object (could be a numpy array for instance)

brandongreenwell-8451:
And the DCO check passed...🥳

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -2145,6 +2145,129 @@ def monotonize(self, term, increasing="auto", passthrough=0.0):
...
+        Returns:
+            Itself.
+        """
+        check_is_fitted(self, "has_fitted_")
+
+        # If term_list contains term names, convert them to indices
+        term_list = [self.term_names_.index(term) if isinstance(term, str) 
+                    else term for term in term_list]
+
+        # Copy any fields we'll overwrite in case someone has a shallow copy of self
```
paulbkoch, line 2167: These copies are not required in this case since we later use a list comprehension which will create a new list.

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -2145,6 +2145,129 @@ def monotonize(self, term, increasing="auto", passthrough=0.0):
...
+        change that attribute directly if desired, and 2) reweighting specific term 
+        contributions will also reweight their related components in a similar 
+        manner (e.g., variable importance scores, standard deviations, etc.).
+
+        Args:
+            weights: A list of weights (one weight for each term in the model). 
+                This should be a list or numpy vector (i.e., 1-d array) of 
+                floats whose i-th element corresponds to the i-th element of the
+                ``.term_*_`` attributes (e.g., ``.term_names_``).
+            remove_nil_terms: Logical indicating whether or not to automatically
```
paulbkoch, line 2210: Boolean instead of Logical in python  ;)

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -2145,6 +2145,129 @@ def monotonize(self, term, increasing="auto", passthrough=0.0):
...
+            raise ValueError(msg)
+
+        # Copy any fields we'll overwrite in case someone has a shallow copy of self
+        term_features = self.term_features_.copy()
+        term_names = self.term_names_.copy()
+        term_scores = self.term_scores_.copy()
+        bagged_scores = self.bagged_scores_.copy()
+        bin_weights = self.bin_weights_.copy()
+        standard_deviations = self.standard_deviations_.copy()
+
```
paulbkoch, line 2231: We're only modifying term_scores, bagged_scores, and standard_deviations, so we don't need copies of the others

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -2145,6 +2145,129 @@ def monotonize(self, term, increasing="auto", passthrough=0.0):
...
+        for idw, w in enumerate(weights):
+
+            scores = term_scores[idw].copy()
+            bscores = bagged_scores[idw].copy()
+            stdevs = standard_deviations[idw].copy()
+            # FIXME: Does it make sense to omit the missing and unknown bins (first
```
paulbkoch, line 2237: For reweighing we should do it for all bins, so we don't need the slice of "1,-1"

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -2145,6 +2145,129 @@ def monotonize(self, term, increasing="auto", passthrough=0.0):
...
+        self.term_names_ = term_names
+        self.term_scores_ = term_scores
+        self.bagged_scores_ = bagged_scores
+        self.standard_deviations_ = standard_deviations
+        self.bin_weights_ = bin_weights
+
+        # Delete "nil" terms (i.e., terms providing zero contribution to the fit)
+        if remove_nil_terms:  # should automatically catch zero weight terms
+            # Grab indices of terms with zero contribution
+            term_list = np.where(self.term_importances() == 0)[0].tolist()
```
paulbkoch: Instead of using term_importances, use weights parameter (and accept any enumerable)

paulbkoch (changes requested):

paulbkoch:
Thanks @brandongreenwell-8451 -- Looks pretty good!  I noted a few small things to change.  Also, please pip install the black formatter, and run "black ." on the command line to pass the final black formatting check in the build.

paulbkoch:
Also, could you add a unit test for each of these in:

https://github.com/interpretml/interpret/blob/develop/python/interpret-core/tests/glassbox/ebm/test_ebm.py

paulbkoch:
I had an additional thought over the weekend regarding naming.  I think for someone new to InterpretML who has not read the documentation, reweigh_terms might be misinterpreted as weighing/measuring something within the EBM rather than something that applies a multiplication factor.  We also use "weight" inside the EBM for the bin_weights_ attribute which might lead someone to guess they are related.  The alternative name I came up with was:

def scale_terms(self, factors, remove_nil_terms=False):

brandongreenwell-8451:
> I had an additional thought over the weekend regarding naming. I think for someone new to InterpretML who has not read the documentation, reweigh_terms might be misinterpreted as weighing/measuring something within the EBM rather than something that applies a multiplication factor. We also use "weight" inside the EBM for the bin_weights_ attribute which might lead someone to guess they are related. The alternative name I came up with was:
> 
> def scale_terms(self, factors, remove_nil_terms=False):

That seems reasonable to me! I'll make that change later this week when I work on your other suggestions.

brandongreenwell-8451:
@paulbkoch I think this addresses all your concerns/suggestions if you want to take another look.

paulbkoch:
Looks great!  Thanks for contributing this @brandongreenwell-8451

bgreenwell:
Awesome @paulbkoch, thanks for the support I’m getting this through!

PR #496, opened 2024-01-09
BUG: define missing logger objects

DerWeh:
There are missing `_log` objects in some modules, which causes errors.


paulbkoch:
Thanks for identifying the missing imports and submitting this PR @DerWeh

PR #503, opened 2024-01-29
MAINT: simplify term_importances

DerWeh:
I found the function `term_importances` unnecessarily complicated. Within the loop, there are 3 branches. To my mind, the function is more understandable if we simply pull out the if clauses.

This handles the pathological case of having only a single class for `importance_type="min_max"` by an early return.

I dropped the use of [itemset](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.itemset.html), as its use is discouraged in the documentation. It makes the code harder to read, and I cannot imagine a case where its performance matters.


paulbkoch:
Thanks @DerWeh -- Nice rewrite!  Looks a lot cleaner now.

PR #511, opened 2024-02-20
Fix minor difference between docstring and actual function

alvanli:
In the [docs](https://interpret.ml/docs/python/api/ExplainableBoostingClassifier.html#interpret.glassbox.ExplainableBoostingClassifier.sweep), the sweep function is supposed to return itself, but none is returned. Added a return statement


paulbkoch:
Nice catch.  Thanks @alvanli!

PR #514, opened 2024-02-26
MAINT: use pydantic for parameter validation

DerWeh:
This is a first proof of concept to use [pydantic](https://docs.pydantic.dev/latest/) for the validation of parameters. I used Python 3.10 syntax as it is more concise and used `pydantic=2.6.2` without putting much thought into it.

Currently, the `fit` methods is extremely messy due to all the checks. Of course, a refactoring putting this into a dedicated method would already simplify things.
I think using `pydantic` might be a more elegant solution. This draft is mostly for discussion, whether this is a good idea or not. I tried to be faithful to the checks that were implemented.
A disadvantage is that `pydantic` errors are harder to read. On the plus side, the type annotations are much clearer than having to search through the fit method. It's probably also much easier to maintain the validation.

Currently, I break the private versions of the classifier. Attributes were set, depending on whether the estimator is private or not. Do you think it would be better to define a super class with the common arguments and subclasses with the additional arguments?

Another question is how to best proceed with overwriting the default arguments. Currently, there is a lot of redundancy. 

If we want to follow through with this idea, I think some sklearn checks should be enabled again using [parametrize_with_checks](https://scikit-learn.org/stable/modules/generated/sklearn.utils.estimator_checks.parametrize_with_checks.html) or [check_estimator](https://scikit-learn.org/stable/modules/generated/sklearn.utils.estimator_checks.check_estimator.html)


paulbkoch:
Hi @DerWeh -- I like what you're doing here and plan to merge it once it fully replaces the existing checks and passes the tests.

A few thoughts:
I think a good minimum version that we should target would be python 3.8 since that's still supported.  Thankfully 3.8 supports things like the Literal annotation.

We can add some classes to formalize the DP vs Regular EBM class partition.  Someday, I think we might support multi-output EBMs, and one strange type of EBM might be an EBM that has both classification and regression outputs.  To create one of the strange beasts, you'd need to directly create an EBMModel object, so I like having the ability to make and use these classes directly.  I can't think of any good reason though to mix private and non-private EBMs, so the private vs regular partition can be a first split in the class hierarchy below the EBMModel class.

You probably already saw that I had a test for check_estimator, but disabled it.  I think we can probably now increase our minimum scikit-learn version to allow for this check now.  We didn't pass all the checks since EBMs are more permissive in what they accept than scikit-learn, and I'd like to preserve our user-friendliness in that respect, so I expect we'll want to continue disabling some of the scikit-learn checks.

DerWeh:
> I think a good minimum version that we should target would be python 3.8 since that's still supported. Thankfully 3.8 supports things like the Literal annotation.

Sure enough, I typically orient myself at [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html), but it is a good idea to be a more conservative (after all, now NumPy is already very stable such that using an older NumPy version is typically sufficient).
I just wanted to avoid the additional boilerplate for type hints for the prototype, I case we reject the idea altogether.

> Someday, I think we might support multi-output EBMs, and one strange type of EBM might be an EBM that has both classification and regression outputs.

Out of curiosity: what do advantage do you expect? I don't really see the synergy for multi-output in EBMs compared to deep networks (or deep trees). Can we expect an advantage during the boosting phase?
But this is mostly an academic question. In practice, I am currently mostly interested in more convenient multi-class support. Interactions and post-processing seem to still be lacking.

> You probably already saw that I had a test for check_estimator, but disabled it.
Yes, I saw it. The PR for the non-private EBMs is ready.

As soon as it is merged, (and when I find time) I can continue working on `pydantic`.

paulbkoch:
> > Someday, I think we might support multi-output EBMs, and one strange type of EBM might be an EBM that has both classification and regression outputs.
> 
> Out of curiosity: what do advantage do you expect? I don't really see the synergy for multi-output in EBMs compared to deep networks (or deep trees). Can we expect an advantage during the boosting phase? But this is mostly an academic question. In practice, I am currently mostly interested in more convenient multi-class support. Interactions and post-processing seem to still be lacking.

I think you're right that there will be very little technical benefit.  Mostly, I think of it as potentially a nice packaging solution for grouping related predictions, although I might reject the idea if it adds anything more than a little complexity.  EBMs do have some benefits in terms of the compactness of the model representation.  For example, SoftAtHome, which wrote Ebm2Onnx (https://github.com/interpretml/ebm2onnx) uses EBMs inside their routers, and I also saw a paper where some company was evaluating EBMs to be used inside pacemakers (!!!).  Sharing the feature definitions, and bin cuts might be nice in those applications both for the compactness of the model, because it could lower the computation requirements of predictions, and also to speed up predictions.

paulbkoch:
> But this is mostly an academic question. In practice, I am currently mostly interested in more convenient multi-class support. Interactions and post-processing seem to still be lacking.

We don't allow multi-class interactions in the default API entirely because of the difficulties in visualizing them.  You can construct multi-class EBMs with interactions today if you use the measure_interactions function manually.  Here's a slightly related example showing how to use it:

https://interpret.ml/docs/python/examples/custom-interactions.html

paulbkoch:
Hi @DerWeh -- I've been thinking through this question the last few days and here's what I came up with in terms of a hierarchy:

EBMBase  -> abstract base class for all EBM models
EBMModel(EBMBase)  -> instantiable class for normal EBM classifiers/regressors
ExplainableBoostingClassifier(EBMModel)
ExplainableBoostingRegressor(EBMModel)
DPEBMModel(EBMBase)  -> instantiable class for differential privacy classifiers/regressors
DPExplainableBoostingClassifier(DPEBMModel)
DPExplainableBoostingRegressor(DPEBMModel)

Thoughts?

DerWeh:
Yes, the hierarchy sound very reasonable. Though, I am not quite sure if making EBMBase and abstract base class has any benefit. See the discussion: https://stackoverflow.com/questions/3570796/why-use-abstract-base-classes-in-python

But it doesn't really seem very important to me, so I am fine with anything/

-----

However, now that EBMs have the checks for `scikit-learn` compatibility, I ran into a bit of a problem: `clone` doesn't work with parameter validation anymore. `clone` performs explicit `is` checks ('old_parameter is new_parameter`) which fail due to `pydantic`...
`pydantic` doesn't offer the option to defer the validation, see the rejected request https://github.com/pydantic/pydantic/issues/559. I knew, that `scikit-learn` urges you to put the validation into the fit method (https://scikit-learn.org/stable/developers/develop.html#instantiation) but I didn't think they are this pedantic. Currently, the only option I see to get through with this PR, is providing our one custom implementation for clone, by providing as custom `_clone_parametrized`, but this seems somewhat ugly...

What is your opinion @paulbkoch?

paulbkoch:
> Yes, the hierarchy sound very reasonable. Though, I am not quite sure if making EBMBase and abstract base class has any benefit. See the discussion: https://stackoverflow.com/questions/3570796/why-use-abstract-base-classes-in-python
> 
> But it doesn't really seem very important to me, so I am fine with anything/

You're probably right that it has very little technical benefit in python.  I was thinking of it mainly as a form of documentation to indicate that the EBMBase class should not be instantiated directly and then just using duck typing as we currently are.  I haven't written as much python as you have (seemingly) and have never made an abstract class in python, but I'm assuming adding the abstract part is fairly lightweight.

> However, now that EBMs have the checks for `scikit-learn` compatibility, I ran into a bit of a problem: `clone` doesn't work with parameter validation anymore. `clone` performs explicit `is` checks (`old_parameter is new_parameter`) which fail due to `pydantic`... `pydantic` doesn't offer the option to defer the validation, see the rejected request [pydantic/pydantic#559](https://github.com/pydantic/pydantic/issues/559). I knew, that `scikit-learn` urges you to put the validation into the fit method (https://scikit-learn.org/stable/developers/develop.html#instantiation) but I didn't think they are this pedantic. Currently, the only option I see to get through with this PR, is providing our one custom implementation for clone, by providing as custom `_clone_parametrized`, but this seems somewhat ugly...
> 
> What is your opinion @paulbkoch?

Yeah, that does seem a bit ugly.  I didn't realize pydantic was unable to use deferred checks.  Even if we get the clone function working though, it's not clear if the scikit-learn requirement to defer checks will break anything else now or in the future.

Maybe we should instead move back to option #2 @DerWeh where we move the checks into a separate _validate function?  That would at least move them out of the way so that someone reading the fit function would only need to see them if they were specifically interested in the checks.  And we'd get to keep the more descriptive error messages too.

paulbkoch:
> Sure, having a separate `_validate` functions is already a first good step. A question to consider is if it is still a good idea to separate EBMs and the DB versions (using your suggested class hierarchy), and if we use the vanilla `dataclass` offered by Python. It doesn't do any validation, but avoids the boilerplate for writing the `__init__` function with all the `self.attribute = attribute` statements.

Avoiding boilerplate through dataclass sounds like a good idea.  I guess this implies having the class hierarchy since we conditionally assign instance attributes based on whether the class is differentially private or not.

> Another question is, if I should keep the ['Yoda-conditions'](https://github.com/MartinThoma/flake8-simplify?tab=readme-ov-file#sim300) that are currently used for type-checks (example: `passthrough < 0.0 or 1.0 < passthrough`). Is this the code-style you use? I don't really know why, but my brain doesn't seem to like these statements, and I always have to mentally convert them to `variable < constant`. Of course, this is a very specific example, where we could use Pythons comparisons `not (0.0 <= passthrough <= 1.0)` (for some reason this is easy to read for me, but the single `0.0 <= passthrough` takes time to digest).

If you want to change the ordering, go for it!  I do use Yoda-conditionals in C++ to combat assignment bugs since "if(a = 5)" compiles, but "if(5 = a)" does not, but I don't care about the order in python since assignment isn't legal inside a conditional in python.

> Anyway, I'll play around a bit more in case I find a reasonable (elegant) way to use `pydantic` for validation. If I don't find a way, I'll at least do the refactor.

Sounds good.  Thanks again for investigating all these options!

DerWeh:
Ok, the parameter validation seems to be working now. I managed to find a way based on pydantics [TypeAdapter](https://docs.pydantic.dev/latest/api/type_adapter/#pydantic.type_adapter.TypeAdapter).

The DP versions are currently broken, as I focused only on the non-private versions.

I am currently looking into breaking up the `fit` method a bit to look for away to avoid as much code duplications as possible when splitting between DP and non-private EBMs.
But this will probably take quite some time. The `fit` methods is quite the monolith, so it takes a lot of time to understand what's going on.

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -219,6 +248,7 @@ def visualize(self, key=None):
             )
 
             return figure
+        # FIXME
```
DerWeh, line 251: If `explanation_type` is neither `"global"` nor `"local"` we end the function without returning. It is bad practice to return only optionally. Should maybe `None` be returned or a `NotImplementedError` or `RuntimeError` be raised?

paulbkoch: I agree.  It should probably raise a ValueError.

DerWeh:
@paulbkoch While the code is not nearly as elegant yet as I would like it to be, this is a first working version of the PR. It would be great if you could help me out with the last failing test. The test `tests/test_explainers.py::test_spec_synthetic` fails for `explainer_class = <class 'interpret.greybox._shaptree.ShapTree'>`. I don't see how this relates to my changes in the EBM. Can you give me a hint?

If we close the pending reviews, I think this could in principle be merged. But I am happy to incorporate feedback. Like I said, it's still quite a step till EBM has a nice readably code. But eliminated 400 lines is a first step.

------
On a related note: it seems to me that the CI tests only Python versions from 3.9 till 3.11. In way opinion, at least the oldest and newest versions should be tests (ideally also for the dependencies). So I would propose to test against Python 3.8.

I also recommend using the linter [ruff](https://docs.astral.sh/ruff/) for code quality of the Python code. By now, it contains a large fraction of checks of other tools, making many obsolete. You have to add a configuration enabling additional rules to get the most of it. It enables integration with [pre-commit](http://pre-commit.com/), a language-server, and a VS Code plugin, making it quite simple to use.

paulbkoch:
Thanks @DerWeh -- I'll take a look at the changes.  -400 lines is great 👍

On the test failure, it looks like it might be due to the 0.0.45 release of Shap that occurred 2 days ago.  The same test is failing on our develop branch, so it isn't likely due to anything you've done.

ruff and tests for 3.8 all sound good.

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -273,95 +303,208 @@ def _clean_exclude(exclude, feature_map):
...
+        Annotated[float, Field(ge=0.0, lt=1.0)],
+    ] = 0.15
+    outer_bags: Annotated[int, Field(ge=1)] = 14
+    inner_bags: NonNegativeInt = 0
+    # Boosting
+    learning_rate: PositiveFloat = 0.01
+    greediness: NonNegativeFloat = 1.5
+    cyclic_progress: Annotated[float, Field(ge=0.0, le=1.0)] = 1.0
+    smoothing_rounds: NonNegativeInt = 200
+    interaction_smoothing_rounds: NonNegativeInt = 50
```
paulbkoch, line 338: The following attributes do not make sense on the DPEBM class and should only exist on the regular EBM class: max_interaction_bins, interactions, inner_bags, greediness, cyclic_progress, smoothing_rounds, interaction_smoothing_rounds, early_stopping_rounds, early_stopping_tolerance, min_samples_leaf, min_hessian

I think to do this we'll need to have separate a separate non-DP EBM class (EBMModel) and the base class for all the attributes which are shared between the DP and non-DP classes (EBMBase).

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -273,95 +303,208 @@ def _clean_exclude(exclude, feature_map):
...
+    term_scores_: List[np.ndarray] = field(init=False)  # np.float64, [bin0..., ?class]
+    standard_deviations_: List[np.ndarray] = field(init=False)  # np.float64, [bin0..., ?class]
+    link_: str = field(init=False)
+    link_param_: float = field(init=False)
+    bag_weights_: np.ndarray = field(init=False)  # np.float64, 1D[bag]
+    breakpoint_iteration_: np.ndarray = field(init=False)  # np.int64, 2D[stage, bag]
+
+    histogram_edges_: List[Union[None, np.ndarray]] = field(init=False)  # np.float64, 1D[hist_edge]
+    histogram_weights_: List[np.ndarray] = field(init=False)  # np.float64, 1D[hist_bin]
+    unique_val_counts_: np.ndarray = field(init=False)  # np.int64, 1D[feature]
```
paulbkoch, line 370: These three attributes should not exist in the DPEBMModel class.

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -273,95 +303,208 @@ def _clean_exclude(exclude, feature_map):
...
+
+    intercept_: np.ndarray = field(init=False)  # np.float64, 1D[class]
```
paulbkoch, line 372: intercept_ is a float on the Regressors (requirement by scikit-learn)

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -273,95 +303,208 @@ def _clean_exclude(exclude, feature_map):
...
+                    stacklevel=2,
+                )
+                return False
+        elif len(self.interactions) == 0:  # interactions must be a list of the interactions
+            return False
+        return True
+
+    def _privacy_parameters(self, term_features, main_bin_weights, sample_weight, domain_size):
+        """Mock method for privacy subclasses."""
+        del term_features, sample_weight, domain_size
```
paulbkoch, line 506: Why delete these?

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -750,29 +622,14 @@ def fit(self, X, y, sample_weight=None, bags=None, init_score=None):  # noqa: C9
...
-                    y,
-                    self.validation_size,
-                    bagged_rng,
-                    0 <= n_classes and not is_differential_privacy,
-                )
-            else:
-                bag = bags[idx]
+        rngs = _create_rngs(rng, n_rngs=self.outer_bags)
+        if bags is None:
+            stratified = n_classes >= 0
```
paulbkoch, line 627: stratified = n_classes >= 0

should instead be:

stratified = n_classes >= 0 and not is_differential_privacy

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -905,19 +706,19 @@ def fit(self, X, y, sample_weight=None, bags=None, init_score=None):  # noqa: C9
...
-                    cyclic_progress,
-                    smoothing_rounds,
+                    self.greediness,
+                    self.cyclic_progress,
+                    self.smoothing_rounds,
                     nominal_smoothing,
                     self.max_rounds,
                     early_stopping_rounds_local,
-                    early_stopping_tolerance,
+                    self.early_stopping_tolerance,
```
paulbkoch, line 721: The attributes that didn't originally have self. should not exist on the DPEBMModel objects.

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -1916,8 +1672,8 @@ def monotonize(self, term, increasing="auto", passthrough=0.0):
         if hasattr(self, "classes_"):
             if len(self.classes_) == 1:
                 # monoclassification is always monotonized
-                return
-            elif 2 < len(self.classes_):
+                return  # FIXME
```
paulbkoch, line 1675: Agreed.  It should return self

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -2467,107 +2224,16 @@ class ExplainableBoostingClassifier(EBMModel, ClassifierMixin, ExplainerMixin):
...
-    bins_: List[Union[List[Dict[str, int]], List[np.ndarray]]]  # np.float64, 1D[cut]
-    feature_names_in_: List[str]
-    feature_types_in_: List[str]
-    feature_bounds_: np.ndarray  # np.float64, 2D[feature, min_max]
-    term_features_: List[Tuple[int, ...]]
-    bin_weights_: List[np.ndarray]  # np.float64, [bin0...]
-    bagged_scores_: List[np.ndarray]  # np.float64, [bag, bin0..., ?class]
-    term_scores_: List[np.ndarray]  # np.float64, [bin0..., ?class]
-    standard_deviations_: List[np.ndarray]  # np.float64, [bin0..., ?class]
-    link_: str
```
paulbkoch, line 2481: link_ could be a Litteral, although we might add new possible Litteral values in successive versions.

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -2467,107 +2224,16 @@ class ExplainableBoostingClassifier(EBMModel, ClassifierMixin, ExplainerMixin):
...
-    link_param_: float
-    bag_weights_: np.ndarray  # np.float64, 1D[bag]
-    breakpoint_iteration_: np.ndarray  # np.int64, 2D[stage, bag]
-
-    histogram_edges_: List[Union[None, np.ndarray]]  # np.float64, 1D[hist_edge]
-    histogram_weights_: List[np.ndarray]  # np.float64, 1D[hist_bin]
-    unique_val_counts_: np.ndarray  # np.int64, 1D[feature]
+    objective: Literal["log_loss"] = "log_loss"
```
paulbkoch, line 2227: objective should accept any strings. We parse these in C++ and python is not even aware of all the possible values.

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -2767,107 +2434,17 @@ class ExplainableBoostingRegressor(EBMModel, RegressorMixin, ExplainerMixin):
...
-    bins_: List[Union[List[Dict[str, int]], List[np.ndarray]]]  # np.float64, 1D[cut]
-    feature_names_in_: List[str]
-    feature_types_in_: List[str]
-    feature_bounds_: np.ndarray  # np.float64, 2D[feature, min_max]
-    term_features_: List[Tuple[int, ...]]
-    bin_weights_: List[np.ndarray]  # np.float64, [bin0...]
-    bagged_scores_: List[np.ndarray]  # np.float64, [bag, bin0...]
-    term_scores_: List[np.ndarray]  # np.float64, [bin0...]
-    standard_deviations_: List[np.ndarray]  # np.float64, [bin0...]
-    link_: str
```
paulbkoch, line 2781: same comment as above.  link_ can be a Literal

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -2767,107 +2434,17 @@ class ExplainableBoostingRegressor(EBMModel, RegressorMixin, ExplainerMixin):
...
-
-    histogram_edges_: List[Union[None, np.ndarray]]  # np.float64, 1D[hist_edge]
-    histogram_weights_: List[np.ndarray]  # np.float64, 1D[hist_bin]
-    unique_val_counts_: np.ndarray  # np.int64, 1D[feature]
-
-    intercept_: float
-    bagged_intercept_: np.ndarray  # np.float64, 1D[bag]
-    min_target_: float
-    max_target_: float
+    objective: Literal["rmse"] = "rmse"
```
paulbkoch, line 2437: same comment as above.  objective should be an unrestricted string.  We do not parse these in python.

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -2884,7 +2461,186 @@ def predict(self, X, init_score=None):
...
+    interaction_smoothing_rounds: Literal[0] = 0
+    max_rounds: NonNegativeInt = 300
+    early_stopping_rounds: Literal[0] = 0
+    early_stopping_tolerance: Literal[0] = 0
+    # Trees
+    min_samples_leaf: Literal[0] = 0
+    min_hessian: Literal[0] = 0
+
+    # Overall
+    random_state: Optional[int] = None
```
paulbkoch, line 2494: I see my earlier comment above about having some of these attributes not exist is handled here by deleting these attributes essentially.  It still feels to me that having a separate non-DP class is a more expected way to handle this, and the subtractive approach I think is also more prone to errors if for example there is an accidental mismatch in one of the attribute names.

Thoughts?

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -2997,98 +2753,15 @@ class DPExplainableBoostingClassifier(EBMModel, ClassifierMixin, ExplainerMixin)
...
-    term_scores_: List[np.ndarray]  # np.float64, [bin]
-    standard_deviations_: List[np.ndarray]  # np.float64, [bin]
-    link_: str
-    link_param_: float
-    bag_weights_: np.ndarray  # np.float64, 1D[bag]
-    breakpoint_iteration_: np.ndarray  # np.int64, 2D[stage, bag]
-
-    noise_scale_binning_: float
-    noise_scale_boosting_: float
+    objective: Literal["log_loss"] = "log_loss"
```
paulbkoch, line 2756: objective can be any string, as above.

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -3259,103 +2932,17 @@ class DPExplainableBoostingRegressor(EBMModel, RegressorMixin, ExplainerMixin):
     noise_scale_boosting\\_ : float
         The noise scale during boosting.
     """
+    objective: Literal["rmse"] = "rmse"
```
paulbkoch, line 2935: objective can be any string, as above.

DerWeh:
@paulbkoch I'll answer your comments in detail as soon as I find the time. For now, I just can give a more general remark:
Currently, I kept the EBMModel and introduced a new DPEBMModel subclass, which basically pins the attributes that are unnecessary for DP. I fully agree, that a cleaner approach is to have the class structure you mentioned: a common base class and a non-private and a DP subclass.
The issue is that at the moment, everything is quite mixed in the `fit` method, such that it's no particular easy to separate the parts without error-prone code duplications.

The big question is, if we use such a monkey-patch approach as an intermediate step, followed by additional clean-up PRs, or if we go for a total revamp from the start. Currently, I went for the first version, as the PR is already a huge change, and I am an outsider of the project. I don't really want to mess things up, and incremental merges might be easier to merge as the changes are more traceable.
But if you want to get it 'right' from the start, I can also try to split non-private and DP right in this PR.

paulbkoch:
Thanks for the background @DerWeh.  We can split it up into 2 PRs if that's easier.  The drawback I see is that it temporarily puts the project in a state where we shouldn't put out a release, but I think that's fine in the short term since I get the sense this is something we can plug the hole in quickly.  It shouldn't break any existing code if anyone clones it, so that's ok too.

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
paulbkoch: Here's the list of possible link values.  Most of these are currently not used, but they could be someday:

https://github.com/interpretml/interpret/blob/d5ec3fa04db8e2acc5fc0de6ca00ff396852caec/shared/libebm/DetermineLinkFunction.cpp#L138-L156

paulbkoch: and here is a list of the link functions we currently implement:

https://github.com/interpretml/interpret/blob/d5ec3fa04db8e2acc5fc0de6ca00ff396852caec/python/interpret-core/interpret/utils/_link.py#L11-L16

paulbkoch:
Hi @DerWeh -- I just noticed something a bit concerning.  I downloaded the documentation produced from the PR and the documentation for the ExplainableBoostingClassifier and ExplainableBoostingRegressor is missing.  Not sure what the problem is yet.

The docs are available to download here:
https://dev.azure.com/ms/interpret/_build/results?buildId=552050&view=artifacts&pathAsName=false&type=publishedArtifacts

We use Jupyter Book to make our documentation which sits on top of Sphinx.  The ExplainableBoostingClassifier is documented here:
https://github.com/interpretml/interpret/blob/develop/docs/interpret/python/api/ExplainableBoostingClassifier.ipynb

I'll have a look.

paulbkoch:
On the documentation, I think we might just need to add the sphinx-pydantic extension:

https://sphinx-pydantic.readthedocs.io/en/latest/

paulbkoch:
> It would be great if you could help me out with the last failing test. The test `tests/test_explainers.py::test_spec_synthetic` fails for `explainer_class = <class 'interpret.greybox._shaptree.ShapTree'>`. I don't see how this relates to my changes in the EBM. Can you give me a hint?

Hi @DerWeh -- I've resolved the ShapTree test that was failing on the develop branch.  It was due to a change in the SHAP API.

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -3259,103 +2932,17 @@ class DPExplainableBoostingRegressor(EBMModel, RegressorMixin, ExplainerMixin):
...
-        random_state: Optional[int] = None,
-        # Differential Privacy
-        epsilon: float = 1.0,
-        delta: float = 1e-5,
-        composition: str = "gdp",
-        bin_budget_frac: float = 0.1,
-        privacy_bounds: Optional[
-            Union[np.ndarray, Mapping[Union[int, str], Tuple[float, float]]]
-        ] = None,
-        privacy_target_min: Optional[float] = None,
```
paulbkoch, line 3323: Is there any way to preserve the parameter ordering of the \_\_init\_\_ functions if some of the attributes are defined in the EBMBase, EBMModel, and ExplainableBoostingClassifier/ExplainableBoostingRegressor classes while still using the dataclass decorators?  It looks by default like the parameters will be grouped by inheritance ordering, so all the EBMBase attributes would come first, then all the EBMModel attributes, etc..

DerWeh:
> Hi @DerWeh -- I just noticed something a bit concerning. I downloaded the documentation produced from the PR and the documentation for the ExplainableBoostingClassifier and ExplainableBoostingRegressor is missing. Not sure what the problem is yet.
> 
> The docs are available to download here: https://dev.azure.com/ms/interpret/_build/results?buildId=552050&view=artifacts&pathAsName=false&type=publishedArtifacts
> 
> We use Jupyter Book to make our documentation which sits on top of Sphinx. The ExplainableBoostingClassifier is documented here: https://github.com/interpretml/interpret/blob/develop/docs/interpret/python/api/ExplainableBoostingClassifier.ipynb
> 
> I'll have a look.

Sorry, never really used Notebooks to generate a documentation, and I am struggling a bit with the Azure pipeline.
However, when I tried to build the documentation locally, the `ExplainableBoostingClassifier` documentation is empty. Is this your issue? Locally, the problem is that `requirements.txt` pins `typing_extensions==4.5.0`. This is not compatible with `pydantic` causing the building of the documentation to fail. If I update `typing_extensions`, everything seems to be fine.

paulbkoch:
> Sorry, never really used Notebooks to generate a documentation, and I am struggling a bit with the Azure pipeline. However, when I tried to build the documentation locally, the `ExplainableBoostingClassifier` documentation is empty. Is this your issue? Locally, the problem is that `requirements.txt` pins `typing_extensions==4.5.0`. This is not compatible with `pydantic` causing the building of the documentation to fail. If I update `typing_extensions`, everything seems to be fine.
> 
> Where does the pinning come from? Can I just update the package?

Yes, just update the pinning if necessary.  I think this is pinned because Jupyter Book changes format frequently and pinning preserves documentation consistency, but we'll need to update these periodically.

paulbkoch:
If we dropped python 3.8 support, we could eliminate the typing_extensions dependency, right?  Seems like a good tradeoff.

paulbkoch:
I've updated the develop branch pipelines to use python 3.9 for building the docs (and other build tasks).

DerWeh:
Sorry for the wait, didn't manage to spend much time on it the last week.

Indeed, we only use [Annotated](https://typing-extensions.readthedocs.io/en/latest/index.html#typing_extensions.Annotated) from `typing_extensions`, which is available in the standard library [typing](https://docs.python.org/3/library/typing.html#typing.Annotated) `3.9`.

However, the import is hard-coded, so I expect, that `typing_extensions` needs to be installed any which ways. Surprisingly, the docs build locally on my machine using Python3.9 on the `requirements.txt` file.

Just googled a little, if you want to still support Python3.8 and keep the dependencies as slim as possible, there is an option to make [dependencies specific for a python version](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#platform-specific-dependencies).
```python
setup(
    ...,
    install_requires=[
        "typing_extensions;python_version<'3.9'",
    ],
)
```
For course, this would complicate the source-code:

```python
import sys
...
if sys.version_info < (3, 9):
    from typing_extensions import Annotated
else:
    from typing import Annotated
```

Your thoughts @paulbkoch?

----------------

I hope that I can resolve the conflicts with interpret 0.6.0 the next days and continue on this PR.

paulbkoch:
Hi @DerWeh -- I'm not too fussy on this, but I think of the options given I'd have a preference towards keeping things simple and dropping support for 3.8.  Anyone who needs python 3.8 can just stick with interpret 0.6.0 until support for 3.8 is officially ended in a few months.

DerWeh:
@paulbkoch The regular and the DP EBMs are no separated (though it is not quite pretty yet). Maybe this is the right point to reevaluate if this is where we want to go.

If so, I'll try to resolve the conflicts (the public API changed). Probably it's would be best, to get it rapidly into a mergeable form, such that the develop branch doesn't deviate further. It's probably simple to add small changes afterward.

paulbkoch:
Hi @DerWeh!  Great to hear you're still looking at this PR.  I'm still thinking the following class hierarchy would be good to implement:

  EBMBase -> abstract base class for all EBM models
  EBMModel(EBMBase) -> instantiable class for normal EBM classifiers/regressors
  ExplainableBoostingClassifier(EBMModel)
  ExplainableBoostingRegressor(EBMModel)
  DPEBMModel(EBMBase) -> instantiable class for differential privacy classifiers/regressors
  DPExplainableBoostingClassifier(DPEBMModel)
  DPExplainableBoostingRegressor(DPEBMModel)

Looking back at our discussion there were some questions about whether EBMBase should be abstract or not, but I think we were in agreement on the rest?  I'm fine with either abstract or not and would defer to someone (you) with more python experience.  It might make more sense to break this PR into one that handles the class hierarchy change, and the rest of this PR, but do whatever works best for you.

Yeah, sorry about all the changes in that section of code.  I added some new configuration options for greedy boosting that required some reworking there.

DerWeh:
Sorry, @paulbkoch I wasn't clear enough. The class hierarchy is implemented by now:

```python
class EBMBase(ABC, BaseEstimator): ....
class EBMModel(EBMBase): ...
class ExplainableBoostingClassifier(EBMModel, ClassifierMixin, ExplainerMixin): ...
class ExplainableBoostingRegressor(EBMModel, RegressorMixin, ExplainerMixin): ...
class DPEBMModel(EBMBase): ...
class DPExplainableBoostingClassifier(DPEBMModel, ClassifierMixin, ExplainerMixin): ...
class DPExplainableBoostingRegressor(DPEBMModel, RegressorMixin, ExplainerMixin): ...
```

It passes all test, but currently cannot be merged as the API on develop changed. What I would like to ask of you, is whether you could check out the PR and give it a try if you like the changes. I only use EBMs in a limited context, and don't have all intended usages in mind. I would like confirmation whether this PR is going the right direction, of course, critique is always more than welcome. When or if you're OK with the PR, I'll resolve the conflicts with the develop branch.


> Yeah, sorry about all the changes in that section of code. I added some new configuration options for greedy boosting that required some reworking there.

Don't worry, it's not your fault that I am so slow. Every improvement is welcome.

paulbkoch:
@DerWeh -- Overall I really like these changes.  The code appears a lot more modular now, and I see you were able to use the new inheritance structure for good effect by creating private functions that differentiate the work based on whether the model is private or not.  I also really like the declarative data validation that pydantic enables.

One area I'm less sure about is the use of @dataclass to remove the boilerplate \_\_init\_\_ functions.  I like the _**theory**_ behind this, but it appears that this change removes some control over the ordering of parameters.  If I'm reading the docs correctly, you can order the \_\_init\_\_ parameters together within each class, but you can't intermix parameters from base classes and child classes.  We do have a common API that should maintain the position of some \_\_init\_\_ parameters.  I'm less familiar with this aspect of our package since I didn't write this part, but here are the example APIs:

https://github.com/interpretml/interpret/blob/develop/python/interpret-core/interpret/ext/examples.py

EBMs fit into the GlassboxExplainer API, which requires having the feature_names and features_types parameters first.  There might be a way to make this work with @dataclass, but in addition to that, I also like how we are able to group common ideas together in the \_\_init\_\_ function and present them to the user in a certain order, even if most people are using keywords instead of position for specifying these.

Does this make sense?

DerWeh:
Thank you for your valuable feedback @paulbkoch. It's important to get the API right. Personally, I opt for keyword-only arguments and declaring the order of parameters an implementation detail. But if you prefer having positional arguments, of course I will respect that.

The question is, if it's enough to be able to explicitly order the attributes of `EBMModel` and `DPEBMModel` (option 1)? Or do you want to go further and allow for `ExplainableBoostingClassifier` and `ExplainableBoostingRegressor` to have argument orders different for `EBMModel`?

Option 1 should be easily doable in the current form: we drop the `@dataclass` decorator from the `EBMBase`. It is abstract anyway, so there is no need to have an `__init__` function. The subclasses `EBMModel` and  `DPEBMModel` stay `dataclass`es and need to repeat all arguments of `EBMBase` defining the order of attributes.

To my knowledge, option 2 forces us to drop `dataclass` completely, we need to write the boilerplate of `__init__` ourselves, but we have complete freedom concerning the attributes. Probably this could be also done replacing the standard library `dataclass` by the much more powerful [attrs](https://www.attrs.org/en/stable/index.html) package, adding a dependency (Context: `attrs` was the original *inspiration* for `dataclasses` https://www.attrs.org/en/stable/why.html#data-classes).
I have no experience with [attrs](https://www.attrs.org/en/stable/index.html), and cannot say if it would be worth adding the dependency.

Are you fine with option 1, or would you prefer option 2, @paulbkoch?

paulbkoch:
Hi @DerWeh -- I'm probably being a bit too sentimental towards compiled languages, but I'd prefer to go with option 2 and keep the \_\_init\_\_ functions.  To me the bigger win in this PR is eliminating all the boilerplate parameter checking because I think that is a bigger barrier to entry for new people coming to the code since new developers have to initially at least skim through most of it to see if has any important logic, which it does!  The \_\_init\_\_ functions in contrast, I feel like people coming to the code can skip over them for the most part after a few seconds because the \_\_init\_\_ is pretty clearly just passing the parameters to the parent class without modifying them.  So, the main pain point with having boilerplate \_\_init\_\_ functions is just maintaining their consistency.  I think we're pretty much done making huge changes to these parameters though.  I can see us adding a parameter here or there, but I don't see that as a maintenance deal-breaker.

Perhaps if we moved the ExplainableBoostingClassifier and ExplanableBoostingRegressor and the DP equivalents into their own files it would further reduce the cognitive load by reducing the length of the _ebm.py file?  I think once people skim over these files, they pretty much don't need to re-examine them.  I've been wanting for some time to move the DP classes into the "privacy" directory (https://github.com/interpretml/interpret/tree/develop/python/interpret-core/interpret/privacy) and it feels like your other changes to separate DP from non-DP make this possible now.  99% of our users don't need the additional cognitive load of DP, so better to move it out in my opinion.

DerWeh:
Sorry for taking such a long break @paulbkoch.
It seems like I won't be able to complete this PR. If I don't at least use a [dataclass](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass), [pydantic.TypeAdapter](https://docs.pydantic.dev/latest/concepts/type_adapter/) doesn't seem to play nicely. Surely, the goals could be achieved somehow, but the solution would probably depend on intrinsic of `pydantic` which is not desirable (brittle and overcomplicated). (My hope that we could also rely on Pydantic for JSON serialization is surely dead.)

I tried to play with [pydantic.validate_call](https://docs.pydantic.dev/latest/api/validate_call/), but that was of no avail, either. I found no nice solution for our object-oriented code.

You can probably close the PR. Sorry for wasting your time.

paulbkoch:
Thanks @DerWeh -- Not a problem at all.  I enjoyed working on this with you and I learned a few things about improved python style that I plan to still incorporate, in addition to the class hierarchy improvements.

paulbkoch:
Closing this PR as requested.

PR #517, opened 2024-02-27
MAINT: Rebase EBM utils

DerWeh:
Simplify the utility functions in the EBM utils.
- reduce code duplication
- reduce complexity by early returns
- explicitly unpack tuples of two elements
- refactor methods
- use dedicated functions of libraries

### DIFF
--- a/python/interpret-core/interpret/glassbox/_ebm/_utils.py
+++ b/python/interpret-core/interpret/glassbox/_ebm/_utils.py
@@ -309,6 +309,8 @@ def deduplicate_bins(bins):
 
 def convert_to_intervals(cuts):  # pragma: no cover
     cuts = np.array(cuts, dtype=np.float64)
+    if cuts.size == 0:
+        return [(-np.inf, np.inf)]
 
     if not np.isfinite(cuts).all():
         raise Exception("cuts must contain only finite numbers")
--- a/python/interpret-core/interpret/glassbox/_ebm/_utils.py
+++ b/python/interpret-core/interpret/glassbox/_ebm/_utils.py
@@ -1,7 +1,8 @@
 # Copyright (c) 2023 The InterpretML Contributors
 # Distributed under the MIT software license
 
-from math import ceil, isnan, isinf, exp, log
+from collections import defaultdict
+from math import ceil, isinf, exp, log, isfinite
 from ...utils._native import Native
 
 from ._tensor import restore_missing_value_zeros
@@ -34,11 +35,11 @@ def convert_categorical_to_continuous(categories):
     # We can't convert a continuous feature that has cuts back into categoricals
     # since the categorical value could have been anything between the cuts that we know about.
 
-    clusters = dict()
+    clusters = defaultdict(list)
     non_float_idxs = set()
 
-    old_min = np.nan
-    old_max = np.nan
+    old_min = +np.inf
+    old_max = -np.inf
     for category, idx in categories.items():
         try:
             # this strips leading and trailing spaces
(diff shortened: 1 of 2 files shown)


DerWeh:
Thank you for the great package. Sadly, some of the high level code is written in a way I found quite complicated. I am currently trying to work through the code, trying to understand what's happening. If you are interested, I would offer to refactor some code on the way, as exemplified by this PR.

paulbkoch:
Your changes seem generally good, and I agree they make the code easier to read.  I'll take a closer look tomorrow.  As you suspect, some of the utils (especially for merging EBMs) are less well tested than they should be.

python/interpret-core/interpret/glassbox/_ebm/_utils.py
```diff
@@ -311,41 +288,34 @@ def deduplicate_bins(bins):
...
-
-    if np.isinf(cuts).any():
-        raise Exception("cuts cannot contain infinity")
+    if not np.isfinite(cuts).all():
+        raise Exception("cuts must contain only finite numbers")
 
-    smaller = np.insert(cuts, 0, -np.inf)
-    larger = np.append(cuts, np.inf)
-    intervals = list(zip(smaller, larger))
+    intervals = [(-np.inf, cuts[0]), *zip(cuts[:-1], cuts[1:]), (cuts[-1], np.inf)]
```
paulbkoch, line 316: This will fail if cuts is empty.  Sorry for not having a test for this.

python/interpret-core/interpret/glassbox/_ebm/_utils.py
```diff
@@ -32,11 +57,11 @@ def convert_categorical_to_continuous(categories):
...
     # since the categorical value could have been anything between the cuts that we know about.
 
-    clusters = dict()
+    clusters = defaultdict(list)
     non_float_idxs = set()
 
-    old_min = np.nan
-    old_max = np.nan
+    old_min = +np.inf
+    old_max = -np.inf
```
paulbkoch, line 64: Using +inf and -inf changes the semantics of the function slightly.  If none of the categories are finite numbers (which would hit the continue statements in the loop below) then this function would return np.nan for the min and max, while here it would return min=+inf and max=-inf

paulbkoch: However, looking at the function a bit longer, I see that I had a bug below in the check "if len(clusters) <= 1" which would subsequently return a single empty array instead of returning "cuts, mapping, old_min, old_max" as is done at the bottom, so the +inf and -inf are not an issue if that return is changed to return np.nan values for the min and max

paulbkoch: If you'd like to fix my bug below please feel free, otherwise I'll correct it after merging this PR.  I think your code above doesn't need any changes if this bug is fixed.

python/interpret-core/interpret/glassbox/_ebm/_utils.py
DerWeh: No problem, it's a `0.` release after all. The clean solution to handle empty cuts is probably an early return:

```python
if cuts.size == 0:
    return [(-np.inf, np.inf)]
```

Is this your expected output for the function? I didn't find any usage of the two methods, so I am not too sure what they are expected to do.

python/interpret-core/interpret/glassbox/_ebm/_utils.py
DerWeh: You are welcome to fix this, as you can probably also add a test-case covering these special cases. I still haven't completely figured out want kind of inputs this function will process.

paulbkoch: Ok, sounds good!

paulbkoch:
Everything other than the minor early return issue looks great!  Thanks so much for simplifying this part of the code @DerWeh!!

DerWeh:
@paulbkoch I added the special case with a minimal test. Good to merge from my side.

paulbkoch:
LGTM.  Merging.

paulbkoch:
And I fixed my bug that didn't handle scenarios where there were no cuts and only 1 interval.

PR #518, opened 2024-02-28
Increase compatibility of EBM with scikit-learn

DerWeh:
This PR adds the estimator checks for scikit-learn compatibility.

We add the following compatibilities:
- add a warning if a 2D y input is used, where we drop the last dimension.
- we do not use a mutable empty list as default argument
- we add a tag that EBMs support NaN inputs in the features
- we check for inf in the classes
- in case of complex inputs we raise an `ValueError` instead of a `TypeError`


DerWeh:
We still fail some scikit-learn tests, where I am not sure if we intentionally deviate.
These are open TODOs/FIXME in `skip_sklearn`.

I didn't remove the old (currently skipped) `test_scikit_learn_compatibility`. Can this test be removed, or is there something important which should be added?

The private versions aren't test at the moment. They fail with some cryptic errors which I do not understand. Sorry, but I don't really understand the DP versions, so I cannot easily fix this part.

I think we need these tests back, before continuing to work on #514, as #514 would mean a rather drastic refactor.

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -551,6 +541,18 @@ def fit(self, X, y, sample_weight=None, bags=None, init_score=None):  # noqa: C9
...
+            # Probably this should be moved. Is hot-fix for sklearn compliance.
+            warn(
+                (
+                    "A column-vector y was passed when a 1d array was"
+                    " expected. Please change the shape of y to "
+                    "(n_samples,), for example using ravel()."
+                ),
+                DataConversionWarning,
+                stacklevel=0,
+            )
```
DerWeh, line 555: This part should most likely be moved to `remove_extra_dimensions`. However, I assume that `clean_dimensions` is not only used for `y`, as you explicitly pass the name to the functions, so it's probably wrong to have the generic check in there.

paulbkoch: Hi @DerWeh -- We actually accept these bad inputs (perhaps controversial) provided we can be reasonably confident that we understand the original intent of the caller.  So, for example if y had the shape (1, 76, 1) and X had the shape (76, 100), that works for us since the caller's intent is clear.  For y we can basically drop any dimension that's of length 1, and if all dimensions are length 1, then y has one sample and 1 feature.  For X, we know how many samples there by processing y first, so we ensure that the number of samples matches up, and we can again drop any extra dimensions that are of length 1 leaving the last dimension that isn't length 1 as the # of features.  I can see why scikit-learn doesn't want all their components to handle these kinds of weird/bad inputs because they'd want agreement between all their components which were written by different people, but we don't have that same issue.

The main place this comes in handy is when someone slices off a column but forgets to remove the extra dimension leading to y having shape=(N, 1)

paulbkoch:
> We still fail some scikit-learn tests, where I am not sure if we intentionally deviate. These are open TODOs/FIXME in `skip_sklearn`.
> 
> I didn't remove the old (currently skipped) `test_scikit_learn_compatibility`. Can this test be removed, or is there something important which should be added?
> 
> The private versions aren't test at the moment. They fail with some cryptic errors which I do not understand. Sorry, but I don't really understand the DP versions, so I cannot easily fix this part.
> 
> I think we need these tests back, before continuing to work on #514, as #514 would mean a rather drastic refactor.

On the scikit-learn tests:
- you're right that we currently don't support sample weights of 0.  There was some issue that I had with them, so I disabled it temporarily.  I think it was a division by zero error if all the samples in a tree leaf had zero weights.  I think this is something that we can eventually support, but I want to be careful about re-enabling it.
- EBMs do support strings.  We support both nominal and ordinal categoricals because we want to be able to show more information in the UI about these kinds of features.  Other packages make you convert a string like "low", "medium" and "high" into 1, 2, 3, but we want the original strings that have more innate human interpretability.
- fitting zero features should be eventually supported, but I'm not sure if it is today. Fitting zero features can be useful if you just want to know what the intercept would be without any features.  For RMSE regression this isn't very useful, but it could be more useful for non-standard objectives.  In any case, I'm not too stressed if we don't support this today as long as we do someday.
- Is the 1.0 != "1.0" error on y values or X values?  We should be treating features with all floating-point numbers as continuous since if we left them as strings then they'd be treated as categoricals, which would be surprising to many people and might not be noticed.  That would make them nominals and would lead to unexpected lower accuracy.
- we do accept 1D X arrays.  I'm a little surprised scikit-learn doesn't too actually for predict since it's handy to slice off a single sample for prediction.  I figure if the user made a mistake and is missing a dimension, then they'll get an exception when the number of samples doesn't match the number in X.

We can remove the old test_scikit_learn_compatibility test that wasn't enabled.

I can have a look at the DP tests once the rest of this PR is in.

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
```diff
@@ -3161,7 +3166,7 @@ class DPExplainableBoostingRegressor(EBMModel, RegressorMixin, ExplainerMixin):
             - `'nominal'`: Categorical where the order has no meaning. Eg: country names. Uses private categorical binning.
     max_bins : int, default=32
         Max number of bins per feature.
-    exclude : list of tuples of feature indices|names, default=[]
+    exclude : list of tuples of feature indices|names, default=None
```
paulbkoch, line 3169: I used an empty array by default here because I felt it was more self-documenting.  If the default is None then you don't know without looking at the documentation what kinds of inputs are expected.  Does a string work?  Does an integer?  Having the empty list at least documents that we're expecting a list of several items.  I'm happy to change this to None though if there's some big issue with scikit-learn compliance or something similar.  The self-documentation aspect is really only minimally useful.

python/interpret-core/setup.py
```diff
@@ -240,7 +240,7 @@ def run(self):
         "numpy>=1.11.1",
         "scipy>=0.18.1",
         "pandas>=0.19.2",
-        "scikit-learn>=0.18.1",
+        "scikit-learn>=0.24",
```
paulbkoch, line 243: Can scikit-learn 0.22 work here, or does 0.24 make the check_estimator significantly nicer?

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
DerWeh: Sure, scikit-learn is a bit more pedantic and issues warnings (which could be filtered). But I agree, it hardly should be an issue. I'll drop the `DataConversionWarning`.

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
DerWeh: I agree that the empty list is a bit more instructive. On the other hand, your type hint `exclude: Optional[Sequence[Union[int, str, Sequence[Union[int, str]]]]] = []` suggests that `None` is a valid value. `Optional` explicitly means `or None`. On the other hand, `exclude=None` is very directly understandable and instructive.

I think the main reason, scikit-learn doesn't allow `exclude=[]`, is that mutable default values are considered extremely harmful in Python. If someone were to modify `exclude`, all following function calls would suddenly have a magically different default values. That's why `None` is the established default value in the Python community. An empty list is very surprising and considered bad practice. Things like `dataclasses` handle this by providing factories instead of mutable default arguments.

python/interpret-core/setup.py
DerWeh: To be honest, I am not sure... The [documentation](https://scikit-learn.org/stable/modules/generated/sklearn.utils.estimator_checks.parametrize_with_checks.html) states `new in 0.24`. But it is not quite clear to me what changed.
Personally, I don't really like supporting `0.` release after the first stable release, but if you consider it important, I can test versions `0.22` till `0.24` and check if there are some issues.

DerWeh:
> - EBMs do support strings. We support both nominal and ordinal categoricals because we want to be able to show more information in the UI about these kinds of features. Other packages make you convert a string like "low", "medium" and "high" into 1, 2, 3, but we want the original strings that have more innate human interpretability.

I updated the tags, indicating this, which allowed re-enabling the test. Please take a look if the tags seem suitable now. Check `ExplainableBoostingClassifier._get_tags()` (and the others) and compare to [the estimator tags](https://scikit-learn.org/stable/developers/develop.html#estimator-tags). We could also use this to skip the tests, but the documentation warns against this.

> - Is the 1.0 != "1.0" error on y values or X values?

The scikit-learn fits floating point `y=[1.0,...]`, and complains that the predictions differ, as EBM predicts string labels `pred=["1.0", ...]`. Thus, I disabled the test. Probably the upstream test should be changed, EBM behavior seems the more reasonable approach to me.
.
> - we do accept 1D X arrays. I'm a little surprised scikit-learn doesn't too actually for predict since it's handy to slice off a single sample for prediction and sometimes people (me especially) forget to keep the same shape. I figure if the user made a mistake and is missing a dimension, then they'll get an exception when the number of samples doesn't match the number in y. Now that I think of it, probably scikit-learn doesn't allow this since they have some methods that only accept an X, so for them it could sometimes be ambiguous if the one dimension was for samples or for features, but for us this isn't an issue since during fitting we always have a y to corroborate the number of samples, or we have a fitted model that can corroborate the number of features.

I guess the ambiguity is whether we have a single sample and the dimension represents features, or a single feature and the dimension represents samples. Anyway, test is skipped.

> We can remove the old test_scikit_learn_compatibility test that wasn't enabled.

Done

> I can have a look at the DP tests once the rest of this PR is in.

Then I'll leave this up to you.

------

I think the PR should be good to merge now. As I said, you can validate that the correct tags are set.

python/interpret-core/interpret/glassbox/_ebm/_ebm.py
paulbkoch: Ok, sounds good.  None it is then.

paulbkoch:
> > * Is the 1.0 != "1.0" error on y values or X values?
> 
> The scikit-learn fits floating point `y=[1.0,...]`, and complains that the predictions differ, as EBM predicts string labels `pred=["1.0", ...]`. Thus, I disabled the test. Probably the upstream test should be changed, EBM behavior seems the more reasonable approach to me. .

Ah, now I remember about this, I converted y floats to strings because JSON doesn't differentiate between integers and floats.  It just has a number type, which are always floats.  I wanted to be able to serialize to JSON and back to a python object.  Without indicating the original datatype it would be ambiguous whether the JSON value 1 should be restored in python as an integer or a float type.  I figured integers would be more common than floats for y, so I converted floats to strings.

If we want to support both integers and floats, we can do it by including an "output_type" field for classifiers in the JSON to differentiate between integers and floats here: https://github.com/interpretml/interpret/blob/0cf3f2f3c4ea7b4cf498e732295defdb712e4ad5/python/interpret-core/interpret/glassbox/_ebm/_json.py#L79

> I think the PR should be good to merge now, or do you want me to check out if we can be more permissive about older scikit-learn versions? As I said, you can validate that the correct tags are set.

If the rest of the PR is ready I can just merge it first and check afterwards.  It's probably fine to use 0.24 but you never know what old environment someone might want to install interpret in, especially a cloud based one where the user might not even be able to upgrade.

I'll review now.

DerWeh:
> If the rest of the PR is ready I can just merge it first and check afterwards. It's probably fine to use 0.24 but you never know what old environment someone might want to install interpret in, especially a cloud based one where the user might not even be able to upgrade.

Just tried to test with an old version, which failed as some test I skip didn't exist in old version... If we compare the functions names instead of functions itself, we can alleviate this problem. 

But more importantly: we actually don't need to bump the version at all, it suffices to increase the scikit-learn version in the `testing` dependencies. It shouldn't be necessary to bother people just using it.

paulbkoch:
Thanks @DerWeh -- It's really nice to finally have scikit-learn verification!

PR #520, opened 2024-03-04
TST: loosen requirements for scikit-learn tests

DerWeh:
This is a fixup for #518 
We only need an up-to-date version for running tests, not for usage of the library.

Furthermore, it is a bit more robust to skip tests based on names instead of comparing functions. This avoids errors, if test aren't available yet in older versions of `scikit-learn`.

---

Sorry for this fixup, these issues slipped my mind when I create the original PR...


paulbkoch:
Thanks @DerWeh, this looks great.

PR #531, opened 2024-04-22
update FAQ on monotonicity

Krzys25:
This PR updates the FAQ part of documentation based on recent changes in how EBMs can handle monotonicity.

The updated response points to two methods of introducing monotonicity and mentions the recommendation of using post-processing (based on the documentation of EBM objects).


paulbkoch:
Thanks @Krzys25!  Nice catch on the update to monotonicity.

Issue #534, opened 2024-05-01
possibility of adding `sample_weight` to `interpret.glassbox.ClassificationTree`


busFred:
the `fit` method of `class sklearn.tree.DecisionTreeClassifier` has two arguments `sample_weight` and `check_input` in addition to the required `X` and `y` `np.ndarray`. I'm wondering if it's possible to just allow additional `sample_weight` to `interpret.glassbox.ClassificationTree.fit` method and pass that `sample_weight` to the underlying `sklearn.tree.DecisionTreeClassifier.fit` method? I'm wondering if the explaining/visualizing procedure requires samples to have uniform weight?

busFred:
Sure, I can definitely do that. seems to be an easy add. However, I do have questions regarding how to set up development environment. I'm wondering what `make install` is doing and why does it need administrator privilege? I'm not that familiar with `nvm` and `npm` and I don't want to do something that could drastically change my work environment and mess other things up. I can work on that once I get clarification.
```shell
(interpret_dev) hungtien@hungtien-T480s-Linux:~/Documents/research/interpret/scripts$ make install
cd /home/hungtien/Documents/research/interpret/shared/vis && npm install

added 1 package, changed 1 package, and audited 1134 packages in 3s

156 packages are looking for funding
  run `npm fund` for details

4 moderate severity vulnerabilities

To address all issues (including breaking changes), run:
  npm audit fix --force

Run `npm audit` for details.
cd /home/hungtien/Documents/research/interpret/shared/vis && npm run build-prod

> @interpretml/interpret-inline@0.6.1 build-prod
> webpack --mode production

asset interpret-inline.js 4.04 MiB [compared for emit] [minimized] [big] (name: main) 1 related asset
orphan modules 339 KiB [orphan] 59 modules
runtime modules 2.62 KiB 8 modules
cacheable modules 9.62 MiB
  modules by path ./node_modules/ 9.32 MiB 26 modules
  modules by path ./src/ 307 KiB
    ./src/index.js + 54 modules 305 KiB [built] [code generated]
    ./node_modules/css-loader/dist/cjs.js!./node_modules/sass-loader/dist/cjs.js!./src/styles.scss 1.38 KiB [built] [code generated]

WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
This can impact web performance.
Assets: 
  interpret-inline.js (4.04 MiB)

[...]

PR #537, opened 2024-05-02
decision tree pass optional kwargs

busFred:
pass optional arguments to `DecisionTreeClassifier.fit` 
#534


paulbkoch:
Thanks @busFred!  This is great  👍

PR #555, opened 2024-06-23
Added APLR algorithm

mathias-von-ottenbreit:
Added glassbox algorithm Automatic Piecewise Linear Regression (and classification).


paulbkoch:
Very nice @mathias-von-ottenbreit!  There's only one thing I would ask before I merge this.  Could you clean out the metadata in the aplr.ipynb file with the command:

```
jupyter nbconvert --to notebook --inplace --ClearOutputPreprocessor.enabled=True --ClearMetadataPreprocessor.enabled=True aplr.ipynb
```

https://github.com/interpretml/interpret/blob/324d5467f2c883cdae71ae31576a48d7f87d6860/docs/interpret/aplr.ipynb

mathias-von-ottenbreit:
Thanks @paulbkoch. I have cleaned the metadata in aplr.ipynb. Thanks for good suggestions yesterday as well. After creating a new python environment I noticed that a newer version of pybind11 was required to build aplr so that it would support numpy 2.0.0.

mathias-von-ottenbreit:
Strange. The only file I changed was aplr.ipynb but some of the tests fail now. This does not seem to be related to aplr though as the following is what fails:

FAILED interpret_core-0.6.2/tests/test_explainers.py::test_spec_synthetic - ValueError: data type <class 'numpy.object_'> not inexact
= 1 failed, 446 passed, 40 skipped, 2 xfailed, 26 warnings in 613.66s (0:10:13)

paulbkoch:
I think it's unrelated since test_spec_synthetic doesn't currently test APLR.  I'll merge this PR and have a look at this test independently on the develop branch.  There are a lot of things breaking this week in many different packages due to the numpy 2.0 release, and this might be another one of them.

paulbkoch:
Not that it matters to this PR anymore, but I think the failure is due to the SciPy release that happened 2 hours ago and is related to this issue:

https://github.com/scipy/scipy/issues/20943

paulbkoch:
Hi @mathias-von-ottenbreit -- I ended up making one addition change after merging, which was to move the APLR package dependency from interpret-core to the interpret package.  The way we've structured our packages is that interpret-core is meant to be a minimal dependency package with no visualization.  The interpret package is our heavyweight, all bells and whistles package that depends on interpret-core, and the other glassbox/blackbox packages.  This architecture doesn't allow our classes to inherit directly from classes in other packages, so we proxy calls to the underlying models.  I've made all the changes required to match how we do this with other packages.  You may want to review my changes in https://github.com/interpretml/interpret/commit/143feefc3d953ad8f8fec6221a3a2c22eebe749a.

PR #560, opened 2024-07-06
APLR

mathias-von-ottenbreit:
Added information about APLR to the readme file (just minimalistic information).

Improved APLR's compatibility with sklearn. This makes it easier to, for example, add APLR to the notebook that produces the benchmarks shown in the readme file. 

If you add APLR to the benchmark notebook at a later stage then I recommend to:
- Set max_interaction_level=0 for datasets where logistic regression does well, such as the Breast Cancer dataset.
- For the larger datasets, significantly increase the learning rate, v, and reduce the maximum number of boosting steps to try, m, in order to speed up the algorithm.


paulbkoch:
Thanks @mathias-von-ottenbreit.

PR #574, opened 2024-09-29
DEV: use pre-commit and ruff to improve Python style

DerWeh:
This PR adds a [pre-commit](https://pre-commit.com/) and [ruff](https://docs.astral.sh/ruff/) configuration to improve the Python programming style.

Exemplarily, I have fixed the issues in `_ebm.py`. The configuration is borrowed from [scientific-python](https://learn.scientific-python.org/development/guides/style/). Please take a look, in case you want to include further checks. Certainly, the configuration can be improved. This is more of proof-of-concept.

To use it locally, you have to use `pre-commit install` after cloning the repo once. Of course, we would like to use it within the CI and not rely on contributors remembering to enable it locally. To my knowledge, the most suitable would be using https://pre-commit.ci/. See the documentation in https://pre-commit.com/#usage-in-continuous-integration.

I would advise to also include checks for the docstrings. However, I could not identify if you follow a certain docstring style. It seems to vary from docstring to docstring.
Personally, I would advise using [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html). But this is mostly a matter of taste.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,32 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: "v4.6.0"
+    hooks:
+      - id: check-added-large-files
+      - id: check-case-conflict
+      - id: check-merge-conflict
+      - id: check-symlinks
+      - id: check-yaml
+      - id: debug-statements
+      - id: end-of-file-fixer
+      - id: mixed-line-ending
+      - id: requirements-txt-fixer
+      - id: trailing-whitespace
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: "v0.6.7"
+    hooks:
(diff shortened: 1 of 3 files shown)


paulbkoch:
Thanks @DerWeh -- I like the idea of moving to ruff, especially now that they also support black style, along with all the other goodies.  There is a black style check in our CI, which I'm sure you've noticed from the one CI failure.  Using pre-commit.ci would be even better.  I tried to do something similar using github actions a while back, but kind of gave up.

In terms of docstrings, I agree on numpydoc.  I transitioned the EBM code to using it but didn't update the rest of the codebase.

DerWeh:
Good to know, then I would add some checks for numpydoc doc strings.

Using pre-commit.ci is straight forward for open source projects, as far as I remember. Feel free to ask me, but I am sure you're much better with  the CI than I am.

Issue #370, opened 2022-09-28
Error in showing global explainations when using more features


R-Z78:
It is very helpful to use Interpret ML which can define explanations. I firstly use 17 features to do prediction via EBM regression, and EBM could show global explainations. However, when I using all the features (i.e., 76 features) to regress, I get the error, "list index out of range".  How can I solve this problem?

IndexError                                Traceback (most recent call last)
Input In [16], in <cell line: 119>()
    115 print('r2 score = %0.2f ± %0.2f' % (mean_r2_score, std_r2_score))
    118 ebm_global = ebm.explain_global()
--> 119 show(ebm_global)
    122 ebm_local = ebm.explain_local(X_test[:5], y_test[:5])
    123 show(ebm_local)

File D:\anaconda\lib\site-packages\interpret\visual\interactive.py:172, in show(explanation, key, **kwargs)
    170 except Exception as e:  # pragma: no cover
    171     log.error(e, exc_info=True)
--> 172     raise e
    174 return None

File D:\anaconda\lib\site-packages\interpret\visual\interactive.py:169, in show(explanation, key, **kwargs)
    166         this.visualize_provider = AutoVisualizeProvider()
    168     # Render
--> 169     this.visualize_provider.render(explanation, key=key, **kwargs)
    170 except Exception as e:  # pragma: no cover
    171     log.error(e, exc_info=True)

File D:\anaconda\lib\site-packages\interpret\provider\visualize.py:224, in InlineProvider.render(self, explanation, key, **kwargs)
    221 def render(self, explanation, key=-1, **kwargs):
    222     from ..visual.inline import render
--> 224     render(
    225         explanation,
    226         default_key=key,
    227         detected_envs=self.detected_envs,
    228         js_url=self.js_url,
    229     )

File D:\anaconda\lib\site-packages\interpret\visual\inline.py:216, in render(explanation, id_str, default_key, detected_envs, js_url)
    214     viz_obj = _build_viz_err_obj(msg)
    215 else:
--> 216     viz_obj = _build_viz_obj(explanation)
    218 init_js, body_js = _build_javascript(
    219     viz_obj, id_str, default_key=default_key, js_url=js_url
    220 )
    222 if "databricks" in detected_envs:

File D:\anaconda\lib\site-packages\interpret\visual\inline.py:100, in _build_viz_obj(explanation)
     98     selector_obj = {"columns": [], "data": []}
     99 else:
--> 100     specific = [
[...]

PR #581, opened 2024-10-24
BUG: Fix index error when values exceed 999T

degenfabian:
Hi!

This PR is an intended fix of the bug presented in issue #370. I fixed the IndexError by falling back to using the scientific notation once values become larger than 999T. I also added a test case which tests that exact functionality and with this modification I was able to solve the IndexError.


paulbkoch:
Thanks @degenfabian.  It's great to have this finally fixed!

Issue #428, opened 2023-04-28
Error: Class Number Mismatch


princessofpillows:
Hi,

I'm experiencing an issue with the interpretml library that I don't understand. Below is an example of the issue.

```
import numpy as np
from lightgbm import LGBMClassifier
from interpret.blackbox import ShapKernel

x = np.random.rand(1000, 20)
y = np.random.randint(low=0, high=1, size=1000)

model = LGBMClassifier()
model.fit(x, y)

shap = ShapKernel(model, data=x)
```

The above code generates this error:

```
Traceback (most recent call last):
  File ".\experiment.py", line 15, in <module>
    shap = ShapKernel(model, data=x)
  File "C:\Users\jopatterson\Documents\SHAP Memory Leak\env_interpretml\lib\site-packages\interpret\blackbox\shap.py", line 44, in __init__
    predict_fn, n_classes, _ = determine_classes(model, data, n_samples)
  File "C:\Users\jopatterson\Documents\SHAP Memory Leak\env_interpretml\lib\site-packages\interpret\utils\_binning.py", line 2833, in determine_classes
    raise ValueError(msg)
ValueError: class number mismatch
```

I'm confused, as replacing interpetml's ShapKernel() with the shap library's Explainer() class works fine.
Does anyone have further insight into this issue?
Thank you.

paulbkoch:
This line generates all zeros:

```
y = np.random.randint(low=0, high=1, size=1000)
```

Since y is always zero, the model should trivially always predict zero, and no features should have any importance since the output does not depend on anything.

It's a bit of a special case, but I agree we should handle this differently.  We should either return zeros for all the importances, or throw an exception.

paulbkoch:
Fixed by @degenfabian in PR #582

PR #582, opened 2024-10-24
ENHANCEMENT: Handle mono classification with Shap

degenfabian:
Hi!
This PR intends to address issue #428. For the edge case described in the issue, instead of throwing an error, the program now outputs a warning that the model was only trained on one class and then creates a wrapper model that assigns 1 to said class and 0 to a synthetic class. This way no error is thrown anymore with the example provided in the issue.


paulbkoch:
Thanks for the fix @degenfabian!  Nice to clear this one out too!

Issue #235, opened 2021-05-18
Use standard deviation from global explanation graph to calculate standard deviation of prediction?


bverhoeff:
(moved this from other issue)
First of all many thanks for creating and maintaining this project.

I have the following related questions about quantifying uncertainty of predictions:

1) Is a standard deviation calculated by bagging data 8 times and training 8 models and averaging them representative for the distribution of uncertainty of the feature scores of a certain feature at certain feature values? (e.g. by calculating a normal distribution with average and SD).

2) Can we 'propagate' this uncertainty towards the final predictions by drawing from the respective abovementioned distributions as apropriate multiple times and calculating predictions multiple times as well, finally resulting in an average and SD of predictions for a certain case.

3) Currently there are no upper and lower bounds for interaction terms in the internal object. Would it be difficult to add these? Otherwise abovementioned method would only be possible without interaction terms

4) Most importantly, would abovementioned method be mathematically correct? I can imagine that this is the wrong place for that question. I could ask this at Stats Exchange / Cross Validated?

code example to illustrate line of thought:

```python
def feature_mean_sd(ebm_global, feature: int, value: float):
    name = ebm_global.__dict__["_internal_obj"]["overall"]["names"][feature]
    if " x " in name:
        return 0, 0, name  # return zeros when interaction, not supported

    feat_dict = ebm_global.__dict__["_internal_obj"]["specific"][feature]
    vals = feat_dict["names"]
    means = feat_dict["scores"]
    upps = feat_dict["upper_bounds"]

    i = np.searchsorted(vals, value)
    if i == len(vals) - 1:  # in case new value above range X_train
        i -= 1
    sd = upps[i] - means[i]

    return means[i], sd, name


def draw_feature_value(ebm_global, feature, value):
    mean, sd, name = feature_mean_sd(ebm_global, feature, value)

    return np.random.normal(mean, sd), name


def pred_with_sd(ebm_global, feature_values):
    samples = []

    for sample in range(16):
        pred = intercept
        for f in range(len(feature_values)):
            pred += draw_feature_value(ebm_global, f, feature_values[f])[0]
        samples.append(pred)
        
    return np.mean(samples), np.std(samples)


ebm_global = ebm.explain_global(name='EBM')
intercept = ebm.intercept_[0]

samples_range, samples_sd = [], []
for i in range(0, 100):
    feature_values = X_test.iloc[i, :].values
    samples = pred_with_sd(ebm_global, feature_values)
[...]

PR #584, opened 2024-10-31
ENHANCEMENT: Add a prediction function that also calculates uncertainty in the prediction

degenfabian:
This PR intends to fix issue #235. Since release v0.3.0 individual bagged models are stored inside the EBM which made it possible to create a prediction function that also calculates uncertainty by calculating the mean and standard deviation across bags (as suggested in the issue #235 by @interpret-ml).

Since this issue was created, some implementation details were changed and I was not able to access the transform function from EBMPreprocessor anymore. Therefore I edited the function construct_bins in _preprocessor.py to return the main_preprocessor and then safe it as class attribute in EBMModel. 

I am not sure if it should be kept this way because an earlier release removed this attribute from EBMModel, but I do not know how else to access the functionality of the transform method (except duplicating it). I would be happy about guidance regarding this.

Another implementation detail which has changed was the structure of bins in the transform function, which threw an error when calling the discretize function. I fixed this by always accessing the last element in the bins list, but I am also happy about guidance here for cleaner or more robust solutions.

Last but not least, this currently only works if max_bins == max_interaction_bins, so at the moment I am just informing the user when calling that these attributes have to be equal.

I also tested the uncertainty estimates on the breast cancer dataset from sklearn and have attached the plot I made to this PR. I think it looks good :)
Figure_1

### DIFF
--- a/python/interpret-core/interpret/glassbox/_ebm/_ebm.py
+++ b/python/interpret-core/interpret/glassbox/_ebm/_ebm.py
@@ -746,7 +746,6 @@ class EBMModel(BaseEstimator):
         missing_val_counts = binning_result[6]
         unique_val_counts = binning_result[7]
         noise_scale_binning = binning_result[8]
-        self._preprocessor = binning_result[9]
 
         if np.count_nonzero(missing_val_counts):
             warn(
@@ -2257,8 +2256,7 @@ class EBMModel(BaseEstimator):
         return self
 
     def pred_from_base_models_with_uncertainty(self, instances):
-        """Gets predictions and uncertainties from the bagged base models.
-
+        """Gets raw scores and uncertainties from the bagged base models.
         Generates predictions by averaging outputs across all bagged models, and estimates
         uncertainty using the standard deviation of predictions across bags.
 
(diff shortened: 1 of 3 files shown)


paulbkoch:
Thanks for this PR @degenfabian.  I do have some suggestions regarding your questions.  There is an internal function called ebm_predict_scores that you can use to simplify this processing.  You can see an example of it being used here:

https://github.com/interpretml/interpret/blob/develop/python/interpret-core/interpret/glassbox/_ebm/_ebm.py#L1453-L1473

and also here: https://github.com/interpretml/interpret/blob/develop/python/interpret-core/interpret/glassbox/_ebm/_ebm.py#L1040-L1049

The ebm_predict_scores function is designed to handle individual bagged models.  To do this you would replace self.intercept_ and self.term_scores_ with slices from self.bagged_intercept_ and self.bagged_scores_.  With this change you shouldn't need the preprocessor anymore, and it should also handle the condition where max_bins != max_interaction_bins.

degenfabian:
Thank you so much for the feedback @paulbkoch, it was really helpful! I tried implementing it to the best of my ability and it looks a lot cleaner now! Do you mind taking another look at it? As you said it, max_bins != max_interaction_bins is not an issue anymore with this updated version!

paulbkoch:
Thanks @degenfabian, the code looks good.  Before I merge though, could you add a test?

degenfabian:
Of course, just added them! So sorry I didn't do that earlier and thank you so much for your valuable and insightful feedback!

paulbkoch:
Yes, this is in the public package:

https://interpret.ml/docs/python/api/ExplainableBoostingClassifier.html#interpret.glassbox.ExplainableBoostingClassifier.predict_with_uncertainty

PR #603, opened 2025-03-14
ENH: speed up _harmonize_tensor

DerWeh:
As far as I can see, cell_map is typically a list with a single tuple. This function call incurs tremendous function overhead for doing nothing (multiplying a single number with 1): first we generate a list, then go through all the NumPy dispatching.

If this indeed is the common case, it's much faster to avoid creating a list (using a generator) and using the simple Python math.prod.

Equivalently, we could use

```python
prod(len(x) for x in cell_map)
```

This is mostly a matter of taste (the performance is identical).

For the simple case of merging a model with itself, this yielded a speed-up of the factor 4 for me. Can you please comment @paulbkoch if my assumptions are correct, or if there are use cases I am overlooking?

------
This speed up is part of PR #578. It doesn't seem that I'll find the time soon to fix the remaining bug in `_harmonize_tensor`. So I thought it's a good idea to include this performance improvement right away.

### DIFF
--- a/python/interpret-core/interpret/glassbox/_ebm/_merge_ebms.py
+++ b/python/interpret-core/interpret/glassbox/_ebm/_merge_ebms.py
@@ -4,7 +4,7 @@
 import logging
 import warnings
 from itertools import chain, count
-from math import isnan
+from math import isnan, prod
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


paulbkoch:
Thanks @DerWeh -- I was wondering why merging EBMs had become so slow in more recent versions.

PR #616, opened 2025-06-15
BUG: fix compatibility with NumPy 2.3

DerWeh:
The C++ data type `BoolEBM` is defined as `int32_t`, thus an integer not a `np.bool_` should be passed to these functions.

NumPy 2.3 removes following behavior:

* ‘np.bool’ scalars can no longer be interpreted as an index (deprecated since 1.19)
  ([gh-28254](https://github.com/numpy/numpy/pull/28254))

I assume that due to this pending deprecation previously it was allowed to pass a `bool` as an `int32_t` as it could be interpreted as index and therefore as number.

-------

Note the subtle difference: Python's `bool` is a `numbers.Integral` and `numbers.Number`:

```Python
>> import numbers
>> isinstance(True, numbers.Integral)
True
>> isinstance(True, numbers.Number)
True
>> True + True
2
```

But `np.bool_` is not:
```python
>> import numpy as np
>> isinstance(np.bool_(True), numbers.Integral)
False
>> isinstance(np.bool_(True), numbers.Number)
False
```

Neither, Python's `bool` nor NumPy's `np.bool_` are NumPy integers or numbers:
```python
>> isinstance(np.True_, np.integer)
False
>> isinstance(np.True, np.number)
False
>> np.True_ + np.True_
np.True_
```

NumPy's `bool` corresponds to `bool` defined in stdbool.h.

------

FYI: as you work with `cytpes`: NumPy offers also some convenience functions to handle pointers and arrays, as well as a platform independent `load_library` implementation.
https://numpy.org/doc/stable/reference/routines.ctypeslib.html

### DIFF
--- a/python/interpret-core/interpret/utils/_compressed_dataset.py
+++ b/python/interpret-core/interpret/utils/_compressed_dataset.py
@@ -6,7 +6,7 @@ import logging
 import numpy as np
 
 from ._clean_x import categorical_encode, unify_columns
-from ._native import Native, _boolebm_t
+from ._native import Native
 
 _log = logging.getLogger(__name__)
 
@@ -81,7 +81,7 @@ def bin_native(
 
         n_bytes += native.measure_feature(
             n_bins,
-            _boolebm_t(np.count_nonzero(X_col) != len(X_col)),
+            np.count_nonzero(X_col) != len(X_col),
             bad is not None,
             feature_type == "nominal",
             X_col,
@@ -134,7 +134,7 @@ def bin_native(
 
         native.fill_feature(
             n_bins,
-            _boolebm_t(np.count_nonzero(X_col) != len(X_col)),
+            np.count_nonzero(X_col) != len(X_col),
             bad is not None,
             feature_type == "nominal",
             X_col,
--- a/python/interpret-core/interpret/utils/_compressed_dataset.py
+++ b/python/interpret-core/interpret/utils/_compressed_dataset.py
@@ -6,7 +6,7 @@ import logging
 import numpy as np
 
 from ._clean_x import categorical_encode, unify_columns
-from ._native import Native
+from ._native import Native, _boolebm_t
 
 _log = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


paulbkoch:
Thanks @DerWeh -- This is great; however I would prefer to convert from bool and np.bool_ into np.int32 inside the _native.py file so that callers to measure_feature and fill_feature do not need to understand these lower-level concepts.

paulbkoch:
C++ has a bool type, but C does not, so the type passed between python and C via ctypes needs to be something like int32.  int32 tends to be the most cross-language compatible type, so that's why it was chosen instead of uint32, int8, int64, etc.  R for example supports int32, but not the others.

The \_native.py file was designed as an interop layer such that code within \_native.py is allowed to do low-level unsafe things and is allowed to look more C-like.  The functions exposed by \_native.py should accept python types.  So, fill_feature and measure_feature should accept python bool types like bool or np.bool_, but not integers.  fill_feature and measure_feature should convert the python bool types they receive into the int32 type expected by the C layer.

I would do something like:

```
if not isinstance(is_missing, (bool, np.bool_)):
    raise ...
is_missing = 1 if is_missing else 0
```

above this (and similar code within \_native.py):
https://github.com/interpretml/interpret/blob/251cf9b8288d32e99f10960532652a8d83c065b4/python/interpret-core/interpret/utils/_native.py#L714-L716

For the ruff error, run:

```
ruff format
```

DerWeh:
@paulbkoch I applied `ruff format` only to the python files, as it would have formatted all notebooks in the documentation folder else. The only file formatted was `python/interpret-core/tests/glassbox/ebm/test_ebm_exact.py` which cause a conflict now.
I assume, the file was formatted on `main` in the meantime. Probably it's best to drop the format commit again and instead rebase against `main`. Or do you want to resolve the conflict?

For the rest, this PR should be good now.

paulbkoch:
The new release just out v0.6.12 has this fix.

PR #631, opened 2025-08-20
MNT: Use bare joblib.Parallel

DerWeh:
Using the `JobLibProvider` adds in my opinion an unnecessary layer of abstracting, increasing complexity. `joblib` is known by many users as is the standard in `scikit-learn`. Providing a custom abstraction makes it more complicated to reasons about the code.

Joblib already offers a backend system:
https://joblib.readthedocs.io/en/stable/custom_parallel_backend.html

For possible `joblib` incompatible extensions, I would argue [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

If we stick with `joblib`, other simplifications would be possible:
In cases like https://github.com/interpretml/interpret/blob/v0.7.2/python/interpret-core/interpret/glassbox/_ebm/_ebm.py#L1287-L1328
we could directly use an iterator instead of building the list and deleting it again.

Furthermore, `joblib` already offers automatic shared memory for multiprocessing:
https://joblib.readthedocs.io/en/stable/parallel.html#working-with-numerical-data-in-shared-memory-memmapping
This is probably sufficient, and could remove the need for manually creating shared memory.

What is your opinion @paulbkoch


ogrisel:
FYI: a fix for the loky issue has been released yesterday in loky 3.5.6 and joblib 1.5.2 (that vendors loky 3.5.6), see https://github.com/joblib/loky/issues/459 and https://github.com/joblib/loky/pull/461 for more details.

Issue #575, opened 2024-09-29
Progress of EBM fitting


DerWeh:
What is the recommended indication for the progress of fitting EBMs? In version 0.5.0, the logger provided the current boosting round and the value of the metric every 10 rounds. This was removed in version 0.6.0. Currently, I find no more progress information of boosting.

For large datasets, fitting an EBM can take several days. So any form of progress indication would be highly welcome. Ideally, we would be also able to save and resume intermediate results (in case of power outages, or we might realize that the results are already good enough for our purpose, or so bad that there is no point in further boosting).

To me, a progress indicator is quite important. It is very hard to estimate the runtime of boosting, as ideally we rely on the early stopping. Thus, the final runtime depends on the “difficulty” of the dataset, making in hard to extrapolate runtimes.

paulbkoch:
A related set of questions has also come up recently in the AutoGluon project with regards to timeouts and callbacks https://github.com/autogluon/autogluon/pull/4480#issuecomment-2365060505, so this would be good for us to solve.  My default thinking until recently has been that other improvements like speed, memory, and model quality were more immediately important, but considering how useful it would be to AutoGluon I now think this is one of the more important things that could be added to EBMs.  Is this something you'd be interested in looking into @DerWeh?

Whatever gets added will need to go somewhere in this loop:
https://github.com/interpretml/interpret/blob/0ad00938c1aab4b06ffd400d358f912d78b54dca/python/interpret-core/interpret/glassbox/_ebm/_boost.py#L83

DerWeh:
Allowing to pass a callback, which can also stop the iterations early, is in principle simple enough and quite general (as long as we define the stopping criteria beforehand). We could either define the important variables which are passed to the callback, or simple pass the locals, which might be more powerful but provides no stable API.

What complicates things is the multiprocessing EBMs default to. Thus, the callback needs to be pickled and runs in a different process...
Is there a reason, a particular reason you stick to multiprocessing instead of multi-threading by default? (See `default="threads"` in [joblib.Parallel](https://joblib.readthedocs.io/en/stable/generated/joblib.Parallel.html#joblib-parallel)) Personally, I observe no big difference in performance. The intensive parts are done in the C++ library, so the GIL is no issue (as long as the library is thread-safe). For a callback, the difference between multi-threading and multiprocessing might be rather significant.
[...]

PR #632, opened 2025-08-23
MNT: Simplify parallel calls

DerWeh:
Instead of generating lists of arguments, dispatching the parallel functions, and then deleting the argument lists, the functions are directly called passing arguments on demand. This should be clearer.
Furthermore, we create a [partial](https://docs.python.org/3/library/functools.html#functools.partial) to avoid repeating common function arguments. This reduces code duplications and makes differences to the boost calls clearer.

---------------------------------------------------

This is a minor refactor to prepare a PR which hopefully eliminates manually managing shared memory. I previously failed with big PRs (#578 isn't abandoned yet, just haven't managed yet to work my way through `_harmonize_tensor`), so I'm trying a more iterative approach now.

### DIFF
--- a/python/interpret-core/interpret/glassbox/_ebm/_ebm.py
+++ b/python/interpret-core/interpret/glassbox/_ebm/_ebm.py
@@ -7,6 +7,7 @@ import logging
 import os
 from copy import deepcopy
 from dataclasses import dataclass, field
+from functools import partial
 from itertools import combinations, count
 from math import ceil, isnan
 from multiprocessing import shared_memory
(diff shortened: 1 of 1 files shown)


DerWeh:
> I already tried the Manager and boolean Value route (see [5f200c8](https://github.com/interpretml/interpret/commit/5f200c8ab197c3d7fcde23d545a96a47a3211c66)). It was A LOT slower. I think the Value uses an OS synchronization primitive, which required locking anytime someone wanted to check the value. With shared memory there isn't any locking required since all threads can just read the boolean value, and the only legal transition for the bool is to signal termination. It would also be possible to pass the boolean into C++ if that became useful, although we don't currently use that.

Thanks for the early feedback. As far as I see, you used a locking value, have you also tried if `lock=False` fixes the performances issues? See [multiprocessing.Value](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Value).

What have you used to check the performance? The tests `test_callbacks_XXX()`? Would be interesting for me to check that my (possible) future simplifications don't hurt performance.


> I did think a bit about the distributed scenario, and I decided the simple option would be to not support the feature that callbacks can terminate all processing nodes. Given that we only parallelize across outer_bags currently, it seems like this wouldn't be a huge limitation in a well-resourced distributed system, because I'd assume processing would start on all outer bags simultaneously.

I agree. Please correct me if I am wrong, but probably the reason for this "synchronized" callback is to create the timeouts when there are fewer jobs than bags (probably hinted in [575](https://github.com/interpretml/interpret/issues/575#issuecomment-2382187333))

The hard part is probably finding out whether we are in a distributed scenario, as the form of parallelism is abstracted away by `joblib`. But I'll look into it once I get there.


> I believe the current code won't work on dask/ray, but I haven't checked. We could make it work in the simple case by simply creating a numpy array instead of using shared memory if the shutdown signal won't be shared. I'm open to other ideas on this. In the ideal world we'd be able to pass a message on termination, but I'm not sure if that's supported in most frameworks since MPI is hard to do properly.

I remember using `dask` as a backend in some previous release, so at one time it worked in principle. (So I have no cluster, so I ran it locally, just using `dask`'s [dashboard](https://docs.dask.org/en/stable/dashboard.html) to inspect the distribution of jobs. This might also be interesting to check the memory usage of EBMs.)

-----------------------------

Thanks for all the useful discussions. I'll include the idea's in the follow-up PR after this one is merged.

paulbkoch:
@DerWeh, I have to admit, I wasn't fond of the current callback mechanism.  I implemented it mainly to get EBMs into AutoGluon where timeouts are either required or highly desirable (https://github.com/autogluon/autogluon/blob/4f492749be1b289333da5d29100985fd5b083f35/tabular/src/autogluon/tabular/models/ebm/ebm_model.py#L19-L30).

Is using dask/ray the only reason to avoid using shared memory?  I've spent quite a bit of time thinking through the parallelism model, and I can't think of a good way to significantly improve parallelization using these frameworks.  There is a way to significantly speed things up using message passing interfaces.  XGBoost and LightGBM seem to have the same problems.  I think the EBM solution will look remarkably similar to how those packages implemented distributed parallelism.

paulbkoch:
The best method I've thought of to massively parallelize the work is to split up the dataset into a number of chunks, then assign each chunk to a processing node, and then compute the histogram statistics on each processing node, followed by a reduce operation to add the histograms from the separate nodes together.  The reduce operation takes log N time, so in theory it should scale to practically any size dataset.  But this will require a fast interconnect between the nodes since we have so many serial boosting steps.  It is essentially the same method employed by XGBoost/LightGBM/catboost.  I've done some of the work to enable this internally, but there's still quite a bit to do.

PR #633, opened 2025-08-23
MNT: use builtin nsmallest

DerWeh:
As far as I can tell, the code simply reimplements the `nsmallest` function available in the standard library.


paulbkoch:
Try changing it to (pair_ranks[idx], idx) to handle ties.

The test could definitely use some documentation.  I added these exact value tests to flag when any real changes in the algorithm occur.  If a change in the code would be expected to change the outputs even slightly, I want to know that, and then update the exact fingerprint if the change in outputs was expected.

PR #636, opened 2025-08-30
MNT: Simplify shared memory

DerWeh:
Use context mangers to handle (shared memory) resources. This avoids to far way `finally` statements for cleanup.


paulbkoch:
Nice.  I didn't like how this looked originally.  This makes it more robust.

PR #638, opened 2025-09-24
Refactor APLR Wrappers for Robustness and Maintainability

mathias-von-ottenbreit:
### Refactor APLR Wrappers for Robustness and Maintainability

This pull request introduces a comprehensive refactoring of the `APLRRegressor` and `APLRClassifier` wrappers located in `interpret/glassbox/_aplr.py`. The primary goal of these changes is to improve the design, robustness, and long-term maintainability of these components by moving from a composition-based pattern to a more elegant and resilient inheritance model.

#### Key Improvements:

1.  **Inheritance over Composition**: The wrappers now directly inherit from their corresponding native `aplr` objects (`APLRRegressorNative` and `APLRClassifierNative`). This change eliminates a significant amount of boilerplate code where methods were simply delegating calls to an internal `self.model_` instance. The new implementation is cleaner, more concise, and better utilizes Python's object-oriented features.

2.  **Safe, Deferred Imports with a Factory Pattern**: The original implementation correctly deferred the import of the `aplr` library to avoid potential side effects at module import time. To preserve this safety while enabling inheritance, this PR introduces a factory pattern.
    *   `APLRRegressor` and `APLRClassifier` are now factory functions that dynamically create the wrapper classes (`APLRRegressorImpl` and `APLRClassifierImpl`) upon their first instantiation.
    *   This ensures the `from aplr import ...` statement is only executed when an APLR model is actually created, completely mitigating the risk of import-time failures.

3.  **Future-Proof `fit` Method**: The signature of the `fit` method in both wrappers has been updated to `fit(self, X, y, **kwargs)`. This makes the wrappers more robust and forward-compatible, as any new parameters added to the native `aplr` library's `fit` method in the future will be automatically passed through without requiring code changes in `interpret-core`.

4.  **Simplified Parameter Handling**: The custom `get_params` and `set_params` methods have been removed. The wrappers now correctly inherit and rely on the scikit-learn-compatible implementations from the parent native `aplr` classes. This reduces code duplication and ensures that parameter handling remains consistent with the underlying library.

5.  **Improved Input Validation**: A bug in the `convert_to_numpy_matrix` utility function has been fixed. It now correctly validates input data, throwing a `TypeError` if a NumPy array or Pandas DataFrame with non-numeric datatypes is provided. This prevents downstream errors and provides clearer feedback to the user on data format issues.

#### Overall Impact:

These changes collectively make the `_aplr.py` module more robust, maintainable, and idiomatic. The code is now better prepared for future updates to the `aplr` dependency and provides a much cleaner integration within the `interpret` ecosystem.

### DIFF
--- a/docs/interpret/aplr.ipynb
+++ b/docs/interpret/aplr.ipynb
@@ -93,7 +93,7 @@
    "source": [
     "<h2>Further Resources</h2>\n",
     "\n",
-    "Below are more detailed code examples from the backing repository for APLR. This will also work when using APLR through `InterpretML` if you change `from aplr import` to `from interpret.glassbox import`.\n",
+    "Below are more detailed code examples from the backing repository for APLR. This will also work when using APLR through `InterpretML` if you change `from aplr import` to `from interpret.glassbox import`. For better precision in global explanation plots you may consider using the plotting functionality in native APLR, introduced in APLR version 10.13.0.\n",
     "- [Regression code example](https://github.com/ottenbreit-data-science/aplr/blob/main/examples/train_aplr_regression.py)\n",
     "- [Classification code example](https://github.com/ottenbreit-data-science/aplr/blob/main/examples/train_aplr_classification.py)"
    ]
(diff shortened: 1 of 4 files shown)


mathias-von-ottenbreit:
Good idea @paulbkoch. I have implemented it.

PR #662, opened 2026-04-12
Fix repeated callback iterations (#635)

### LINKED ISSUE #635: Repeated iterations in callback
I tried using the `callback` to provide progress indicated. But using a simple `print` functions, I am quite confused by the output: Why do I get repeated output for the same step and back, as soon as the progress halts?

Example code:
```python
from interpret.glassbox import ExplainableBoostingRegressor
from interpret.utils import make_synthetic
from sklearn.model_selection import train_test_split

seed = 42

X, y, names, types = make_synthetic(classes=None, n_samples=50000, missing=False, seed=seed)
X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.20, random_state=seed)


def logging_callback(bag_idx: int, n_steps: int, has_progressed: bool, best_score: float) -> bool:
    if n_steps % 10000 == 0:
        print(f"{bag_idx=:2}, {n_steps=:5}, {has_progressed=!s:>5}, {best_score=:.6}")
    return False


ebm = ExplainableBoostingRegressor(names, types, interactions=3, smoothing_rounds=5000, reg_alpha=10.0, callback=logging_callback)
ebm.fit(X_train, y_train)
```

Some output:
```
bag_idx= 6, n_steps=100000, has_progressed= True, best_score=0.313481
bag_idx= 6, n_steps=100000, has_progressed=False, best_score=0.313481
bag_idx= 6, n_steps=100000, has_progressed=False, best_score=0.313481
bag_idx= 6, n_steps=100000, has_progressed=False, best_score=0.313481
bag_idx= 6, n_steps=100000, has_progressed=False, best_score=0.313481
bag_idx= 6, n_steps=100000, has_progressed=False, best_score=0.313481
bag_idx= 6, n_steps=100000, has_progressed=False, best_score=0.313481
bag_idx= 6, n_steps=100000, has_progressed=False, best_score=0.313481
bag_idx= 6, n_steps=100000, has_progressed=False, best_score=0.313481
bag_idx= 6, n_steps=100000, has_progressed=False, best_score=0.313481
bag_idx= 6, n_steps=100000, has_progressed=False, best_score=0.313481
```

ugbotueferhire:
Fixes #635 Repeated Callback Iterations
## Description
This PR resolves an issue during EBM training where the `callback` parameter is invoked repeatedly with the exact same `n_steps` value whenever the boosting loop fails to make progress.
### The Bug
In the boosting loop inside `_boost.py`, the callback was being unconditionally invoked at the very end of the loop, regardless of whether `make_progress` was `True` or `False`. Because `step_idx` is only incremented when `make_progress` is `True`, non-progressing cycle iterations continuously spammed the callback with the exact same `n_steps` value, resulting in repeating redundant output logs for the user.
### The Fix
I moved the callback invocation **inside** the `if make_progress:` block, right after early stopping tolerance evaluations.
```diff
-                        if callback is not None:
-                            is_done = callback(
-                                bag_idx, step_idx, make_progress, cur_metric
-                            )
-                            if is_done:
-                                if stop_flag is not None:
-                                    stop_flag[0] = True
-                                break
```
```python
+                            if callback is not None:
+                                is_done = callback(
+                                    bag_idx, step_idx, make_progress, cur_metric
+                                )
+                                if is_done:
+                                    if stop_flag is not None:
+                                        stop_flag[0] = True
+                                    break
```
This change strictly guarantees that the callback only receives monotonic, progressing `n_steps` values.
## Design Decisions
* The `has_progressed` parameter has been strictly maintained for API backward compatibility. Because the callback is now only ever fired after progress is made, it will always be evaluated as `True`. This ensures any pre-existing user callbacks utilizing `if has_progressed:` do not unexpectedly break.
## Testing
Added 5 regression tests inside a new `test_callback.py` file to maintain the invariant:
1. `test_callback_no_repeated_steps_classifier`: Verifies `n_steps` remains monotonically increasing across distinct boosting phases (main terms vs interaction pairs) without any duplicate step integers.
2. `test_callback_no_repeated_steps_regressor`: Equivalent functionality verification for the regressor implementation.
3. `test_callback_has_progressed_always_true`: Validates `has_progressed` correctly always outputs `True`.
4. `test_callback_early_termination`: Assertions establishing that `is_done = True` continues to properly bypass early stopping settings to instantly conclude training.
5. `test_callback_receives_valid_metrics`: Ensures the `best_score` metric received by the callback does not evaluate to NaN or infinite sequences.
All existing and newly included formatting validation and regression tests successfully pass without issue.


ugbotueferhire:
@paulbkoch Hello again I’ve opened this PR to fix the repeated callback iteration issue (#635).

The callback now strictly fires inside the make_progress block, which completely eliminates the duplicate n_steps spam. Keeping API compatibility in mind, the has_progressed parameter remains in the signature and will always evaluate to True. I've also added 5 regression tests inside test_callback.py to ensure monotonic n_steps behavior and verify that early termination continues to work flawlessly.

Let me know if you’d like any tweaks made to the tests or if anything needs adjustments thanks.

paulbkoch:
Hi @ugbotueferhire -- Thanks for this PR.  I think I'd like to take this API in a slightly different direction though.  The issue is that n_steps encodes valuable information, which is the number of boosting steps that have been taken, and that's useful in terms of evaluating overfitting.  Changing it to be a monotonic increasing value eliminates that signal.  The real problem I've come to realize is that we actually have two types of callbacks that we're invoking here: a callback to indicate boosting progress, and a callback to indicate that a feature was examined to calculate its gain and the potential update associated with that gain.  But we only have 1 callback parameter, and I don't want to pollute the primary API with tons of different callback parameters.  I think the better API change will be to change the callback parameter such that it is required to use named arguments in the signature.  We can then introspect the function passed to us and determine what kind of callback we were given.  Then, after that change is made, we can break the callbacks into two separate kinds of callbacks with different parameter names, and call them independently.  The different callbacks can then be passed in as tuples via the callback parameter.

ugbotueferhire:
> Hi @ugbotueferhire -- Thanks for this PR. I think I'd like to take this API in a slightly different direction though. The issue is that n_steps encodes valuable information, which is the number of boosting steps that have been taken, and that's useful in terms of evaluating overfitting. Changing it to be a monotonic increasing value eliminates that signal. The real problem I've come to realize is that we actually have two types of callbacks that we're invoking here: a callback to indicate boosting progress, and a callback to indicate that a feature was examined to calculate its gain and the potential update associated with that gain. But we only have 1 callback parameter, and I don't want to pollute the primary API with tons of different callback parameters. I think the better API change will be to change the callback parameter such that it is required to use named arguments in the signature. We can then introspect the function passed to us and determine what kind of callback we were given. Then, after that change is made, we can break the callbacks into two separate kinds of callbacks with different parameter names, and call them independently. The different callbacks can then be passed in as tuples via the callback parameter.

alright understood

ugbotueferhire:
@paulbkoch 
Before I start on this, I have a few questions to make sure I get the design exactly right:
1.  What exact parameter names do you want for each callback type? 
   - For example, for the progress callback: `def progress_cb(*, bag_idx, n_steps, best_score):`? 
   - And for the examination callback: `def exam_cb(*, bag_idx, n_steps, term_idx, avg_gain):`? Or different names?
2. The current callback uses positional args `(bag_idx, n_steps, has_progressed, best_score)`. Should we keep supporting the old positional-arg style during a deprecation period, or is a clean break acceptable?
3. You mentioned eventually passing tuples like `callback=(progress_cb, exam_cb)`. Should I implement tuple support in this PR, or just the keyword-arg introspection for now and leave tuples for a follow-up?

paulbkoch:
@ugbotueferhire -- The next interpret release is going to be a major version bump, so right now is a good time to make breaking changes.

Let's go with:

def progress_cb(*, bag, step, term, metric)

With metric being the current metric, not the best metric.  The caller can keep track of the previous/best metric per bag if that want that information, and having the current metric instead of the best allows them to make decisions about when to end boosting if the current metric starts getting worse.

and:

def exam_cb(*, bag, step, term, gain)

Sound good?

Autogluon needs both callbacks in order to terminate boosting on a time schedule, so we'll need to have both at the next release.  I'm fine having them as separate PRs though, or combine them in a single PR -- whatever works for you.  My suggestion though would be to break them into separate PRs and first implement just the args -> kwargs change followed with a subsequent PR to accept tuples of both these types.

ugbotueferhire:
> @ugbotueferhire -- The next interpret release is going to be a major version bump, so right now is a good time to make breaking changes.
> 
> Let's go with:
> 
> def progress_cb(*, bag, step, term, metric)
> 
> With metric being the current metric, not the best metric. The caller can keep track of the previous/best metric per bag if that want that information, and having the current metric instead of the best allows them to make decisions about when to end boosting if the current metric starts getting worse.
> 
> and:
> 
> def exam_cb(*, bag, step, term, gain)
> 
> Sound good?
> 
> Autogluon needs both callbacks in order to terminate boosting on a time schedule, so we'll need to have both at the next release. I'm fine having them as separate PRs though, or combine them in a single PR -- whatever works for you. My suggestion though would be to break them into separate PRs and first implement just the args -> kwargs change followed with a subsequent PR to accept tuples of both these types.

yeah this sounds good to me

ugbotueferhire:
@paulbkoch I just made an update you can check it out when you free

python/interpret-core/interpret/glassbox/_ebm_core/_boost.py
```diff
@@ -364,18 +364,21 @@ def boost(
...
+                                    bag=bag_idx,
+                                    step=step_idx,
+                                    term=term_idx,
+                                    metric=cur_metric,
+                                )
+                                if is_done:
+                                    if stop_flag is not None:
+                                        stop_flag[0] = True
+                                    break
+
```
paulbkoch, line 383: This check should precede the callback.  It's a piece of shared memory between processes that allows any process to shut the rest down.  After one process signals a stop condition, we shouldn't call the callback in the other processes.  We can't guarantee this, but we can make it very unlikely with this check.

paulbkoch:
@ugbotueferhire -- looking at the tests, I also just realized we should probably include a "stage" parameter in the callback where 0=mains, 1=pairs, etc.  Currently, after we finish the mains and move to the pairs the "step" parameter drops down to 0.  Adding a "stage" before it allows the tuple (stage, step) to be monotonically increasing.

ugbotueferhire:
> @ugbotueferhire -- looking at the tests, I also just realized we should probably include a "stage" parameter in the callback where 0=mains, 1=pairs, etc. Currently, after we finish the mains and move to the pairs the "step" parameter drops down to 0. Adding a "stage" before it allows the tuple (stage, step) to be monotonically increasing.

okay sounds great

python/interpret-core/tests/glassbox/ebm/test_callback.py
```diff
@@ -0,0 +1,246 @@
...
+    if not steps:
+        return []
+    phases = [[steps[0]]]
+    for i in range(1, len(steps)):
+        if steps[i] < steps[i - 1]:
+            # step dropped — new phase started
+            phases.append([steps[i]])
+        else:
+            phases[-1].append(steps[i])
+    return phases
```
paulbkoch, line 57: We no longer need _split_into_phases as we can now check if (stage, step) is monotonically increasing.

python/interpret-core/tests/glassbox/ebm/test_callback.py
```diff
@@ -0,0 +1,246 @@
...
+    ebm.fit(X, y)
+
+    assert len(cb.records) > 0, "Callback should have been invoked at least once"
+
+    steps_by_bag = {}
+    for bag, stage, step, _, _ in cb.records:
+        steps_by_bag.setdefault(bag, []).append((stage, step))
+
+    for bag, steps in steps_by_bag.items():
+        for phase in _split_into_phases(steps):
```
paulbkoch, line 90: same as above -- we can eliminate the split into phases and just check that (stage, step) are increasing.

python/interpret-core/tests/glassbox/ebm/test_callback.py
```diff
@@ -0,0 +1,246 @@
...
```
paulbkoch, line 123: same as above -- we can just check that (stage, step) is increasing without the phases.

COMMENT_DRAFT.md
```diff
@@ -0,0 +1,11 @@
+@paulbkoch That makes a lot of sense — separating progress callbacks from feature-examination callbacks is a cleaner design.
```
paulbkoch, line 1: this file can be deleted

EMAIL_DRAFT.md
```diff
@@ -0,0 +1,35 @@
+**To:** interpret@microsoft.com
```
paulbkoch, line 1: this file can be deleted

EMAIL_DRAFT.txt
```diff
@@ -0,0 +1,24 @@
+To: interpret@microsoft.com
```
paulbkoch, line 1: this file can be deleted

PR_DRAFT.md
```diff
@@ -0,0 +1,45 @@
+Fixes #635 Repeated Callback Iterations
```
paulbkoch, line 1: this file can be deleted

paulbkoch (changes requested):

paulbkoch:
Thanks @ugbotueferhire!

ugbotueferhire:
hello @paulbkoch hope you good
regarding Phase 2 of the callback API updates we discussed, which includes:
_Permitting the callback parameter to accept a tuple/list of callbacks.
_Introducing a separate exam_cb parameter to isolate feature examination.

Please let me know if you are ready for me to begin implementation.

paulbkoch:
Yes, that would be great.  Thanks  @ugbotueferhire!

PR #664, opened 2026-04-27
Add license information to setup.py

ecederstrand:
Fixes license specifier according to https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files

This allows automatic license inspection tools like pip-licenses to correctly categorize this package.


paulbkoch:
Thanks @ecederstrand!

PR #665, opened 2026-05-02
feat: add tuple support and exam_cb for callbacks

### LINKED ISSUE #635: Repeated iterations in callback

ugbotueferhire:
Fixes #635 Callback tuple support and exam callback

## Description

This PR implements Phase 2 of the callback redesign for EBM training.

Phase 1 changed the callback API to keyword-only arguments and ensured the
progress callback only fires on progressing boosting steps. That work was
merged in #662.

This PR builds on top of that by allowing `callback` to accept either:

- a single progress callback
- a single examination callback
- a tuple containing one progress callback and one examination callback

The supported callback signatures are:

```python
def progress_cb(*, bag, stage, step, term, metric):
    ...

def exam_cb(*, bag, stage, step, term, gain):
    ...
```

## Why

The original callback parameter was carrying more than one kind of signal.
As discussed in the review for #662, there are really two different callback
use cases:

- a progress callback that reports accepted boosting progress
- an examination callback that reports when a term has been examined and what
  gain was computed for it

Splitting those concepts makes the API clearer while still keeping a single
public `callback` parameter.

## What Changed

- Added callback normalization and validation in `_ebm.py`
- Classified callbacks by signature using `metric` vs `gain`
- Allowed `callback` to be either a callable or a tuple of callables
- Enforced at most one progress callback and one examination callback
- Added an `exam_callback` hook in `_boost.py` at the point where `avg_gain`
  is computed
- Kept the existing progress callback behavior after accepted term updates
- Allowed either callback type to stop training early by returning `True`
- Updated callback parameter docstrings for EBM model, classifier, and regressor

## Validation Rules

This PR raises a `ValueError` for invalid callback configurations such as:

- empty callback tuples
- more than one progress callback
- more than one examination callback
- callbacks that do not match either supported signature

## Tests

Added or covered tests for:

- tuple callback support calling both callback types
- valid finite `gain` values in the examination callback
- early termination from the examination callback
- invalid tuple configurations
- invalid callback signatures

Verified with:

```bash
python -m pytest tests/glassbox/ebm/test_callback.py -q
python -m pytest tests/glassbox/ebm/test_ebm.py::test_callbacks_short tests/glassbox/ebm/test_ebm.py::test_callbacks_long tests/glassbox/ebm/test_merge_ebms.py::test_merge_ebms_callback_is_none -q
```

## Notes

- This preserves the Phase 1 progress callback behavior introduced in #662.
- The final intercept-only pass still does not invoke training callbacks.

### DIFF
--- a/python/interpret-core/interpret/glassbox/_ebm.py
+++ b/python/interpret-core/interpret/glassbox/_ebm.py
@@ -2,6 +2,7 @@
 # Distributed under the MIT software license
 
 import heapq
+import inspect
 import json
 import logging
 import os
@@ -76,6 +77,88 @@ from ._ebm_core._utils import (
 )
 
 _log = logging.getLogger(__name__)
+
+
+_PROGRESS_CALLBACK_NAMES = ("bag", "stage", "step", "term", "metric")
+_EXAM_CALLBACK_NAMES = ("bag", "stage", "step", "term", "gain")
+_CallbackSpec = Callable[..., bool] | tuple[Callable[..., bool], ...]
+
+
+def _classify_callback(callback):
+    if not callable(callback):
+        msg = "callback must be a callable or a tuple of callables"
+        _log.error(msg)
+        raise ValueError(msg)
+
+    try:
+        signature = inspect.signature(callback)
+    except (TypeError, ValueError) as exc:
+        msg = "callback must have an inspectable signature"
+        _log.error(msg)
+        raise ValueError(msg) from exc
+
+    has_metric = "metric" in signature.parameters
+    has_gain = "gain" in signature.parameters
+    if has_metric == has_gain:
+        msg = (
+            "callback must accept either the progress signature "
+            "(*, bag, stage, step, term, metric) or the examination signature "
+            "(*, bag, stage, step, term, gain)"
+        )
+        _log.error(msg)
+        raise ValueError(msg)
+
+    required_names = _PROGRESS_CALLBACK_NAMES if has_metric else _EXAM_CALLBACK_NAMES
+    missing_names = [
+        name for name in required_names if name not in signature.parameters
+    ]
+    if missing_names:
+        msg = f"callback is missing required parameters: {missing_names}"
+        _log.error(msg)
+        raise ValueError(msg)
+
+    try:
+        signature.bind(**{name: None for name in required_names})
+    except TypeError as exc:
+        msg = f"callback must be callable with keyword arguments {required_names}"
+        _log.error(msg)
+        raise ValueError(msg) from exc
+
+    return "progress" if has_metric else "exam"
+
+
+def _normalize_callbacks(callback):
+    if callback is None:
+        return None, None
+
+    callbacks = callback if isinstance(callback, tuple) else (callback,)
+    if len(callbacks) == 0:
+        msg = "callback tuple cannot be empty"
+        _log.error(msg)
+        raise ValueError(msg)
+    if len(callbacks) > 2:
+        msg = "callback tuple can contain at most one progress callback and one examination callback"
+        _log.error(msg)
+        raise ValueError(msg)
+
+    progress_callback = None
+    exam_callback = None
+    for callback_item in callbacks:
+        callback_type = _classify_callback(callback_item)
+        if callback_type == "progress":
+            if progress_callback is not None:
+                msg = "callback tuple cannot contain more than one progress callback"
+                _log.error(msg)
+                raise ValueError(msg)
+            progress_callback = callback_item
+        else:
+            if exam_callback is not None:
+                msg = "callback tuple cannot contain more than one examination callback"
+                _log.error(msg)
+                raise ValueError(msg)
+            exam_callback = callback_item
+
+    return progress_callback, exam_callback
 
 
 class EBMExplanation(FeatureValueExplanation):
@@ -825,7 +908,8 @@ class BaseEBM(LocalExplainer, GlobalExplainer, SKBaseEstimator):
             interaction_smoothing_rounds = 0
             early_stopping_rounds = 0
             early_stopping_tolerance = 0.0
-            callback = None
+            progress_callback = None
+            exam_callback = None
             min_samples_leaf = 0
             min_hessian = 0.0
             reg_alpha = 0.0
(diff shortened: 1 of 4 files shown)


ugbotueferhire:
@paulbkoch conflicts has been resolved

python/interpret-core/interpret/glassbox/_ebm.py
```diff
@@ -78,6 +79,88 @@
...
+
+    callbacks = callback if isinstance(callback, tuple) else (callback,)
+    if len(callbacks) == 0:
+        msg = "callback tuple cannot be empty"
+        _log.error(msg)
+        raise ValueError(msg)
+    if len(callbacks) > 2:
+        msg = "callback tuple can contain at most one progress callback and one examination callback"
+        _log.error(msg)
+        raise ValueError(msg)
```
paulbkoch: We can get rid of these checks.  Empty callback tuple should be allowed and return None, None which it does in the code below already.  And if callbacks has more than two values then it will be detected by the duplicate checks below.

python/interpret-core/interpret/glassbox/_ebm.py
```diff
@@ -78,6 +79,88 @@
...
+
+    try:
+        signature.bind(**{name: None for name in required_names})
+    except TypeError as exc:
+        msg = f"callback must be callable with keyword arguments {required_names}"
+        _log.error(msg)
+        raise ValueError(msg) from exc
+
+    return "progress" if has_metric else "exam"
```
paulbkoch: This is probably simpler and more expandable as something like:

callback_types = {"progress" : {"bag", "stage", "step", "term", "metric"}, "exam": {"bag", "stage", "step", "term", "gain"}}

param_names = set(inspect.signature(callback).parameters)

for name, params in callback_types.items():
    if params == param_names:
         return name

raise something

python/interpret-core/interpret/glassbox/_ebm_core/_boost.py
```diff
@@ -264,6 +265,19 @@ def boost(
                                 # penalize nominals a bit because they benefit from sorting categories
                                 avg_gain *= gain_scale
```
paulbkoch, line 267: should check 

                            if stop_flag is not None and stop_flag[0]:
                                break

here

python/interpret-core/interpret/glassbox/_ebm_core/_boost.py
paulbkoch, line 399: This one can be removed since we're checking before both callbacks

python/interpret-core/interpret/glassbox/_ebm.py
```diff
@@ -3431,13 +3519,13 @@ def __init__(
         # Boosting
         learning_rate: float = 0.02,
         greedy_ratio: float | None = 10.0,
-        cyclic_progress: bool | float = False,
+        cyclic_progress: bool | float | int = False,  # noqa: PYI041
```
paulbkoch: The float type hint includes int, so remove int and the noqa

python/interpret-core/interpret/glassbox/_ebm.py
```diff
@@ -3755,13 +3843,13 @@ def __init__(
         # Boosting
         learning_rate: float = 0.015,
         greedy_ratio: float | None = 10.0,
-        cyclic_progress: bool | float = False,
+        cyclic_progress: bool | float | int = False,  # noqa: PYI041
```
paulbkoch: remove int since float includes the int type hint, and also remove noqa

python/interpret-core/interpret/glassbox/_ebm.py
```diff
@@ -4085,13 +4173,13 @@ def __init__(
         # Boosting
         learning_rate: float = 0.04,
         greedy_ratio: float | None = 10.0,
-        cyclic_progress: bool | float = False,
+        cyclic_progress: bool | float | int = False,  # noqa: PYI041
```
paulbkoch: same as above.  remove int and noqa

python/interpret-core/tests/glassbox/ebm/test_callback.py
```diff
@@ -223,3 +248,273 @@ def __call__(self, *, bag, stage, step, term, metric):
...
+
+@pytest.mark.parametrize(
+    "callback, message",
+    [
+        ((RecordingCallback(), RecordingCallback()), "more than one progress callback"),
+        (
+            (ExamRecordingCallback(), ExamRecordingCallback()),
+            "more than one examination callback",
+        ),
+        (tuple(), "cannot be empty"),
```
paulbkoch: empty tuple should be valid as a no callbacks case

python/interpret-core/tests/glassbox/ebm/test_ebm.py
```diff
@@ -249,6 +249,32 @@ def test_ebm_sweep():
...
+        interactions=interactions,
+        outer_bags=1,
+        max_rounds=10,
+        n_jobs=1,
+    )
+
+    with pytest.raises(ValueError, match=message):
+        ebm.fit(X, y)
+
+
```
paulbkoch: this test seems out of scope for the callback change

paulbkoch:
Looks great!  Thank you @ugbotueferhire.

PR #666, opened 2026-05-09
data: fix Marginal.visualize crash on categorical features (#119)

### LINKED ISSUE #119: visualize has a bug for categorical features
This is regarding this file:
https://github.com/interpretml/interpret/blame/master/python/interpret-core/interpret/data/response.py

MarginalExplanation.visualize(key) has a bug that prevents it from processing categorical features correctly. Changing the following lines will fix it:

move line #137:
            bin_size = density_dict["names"][1] - density_dict["names"][0]

to be in the else (line #142)
        if is_categorical:
            trace1 = go.Histogram(x=data_dict["x"], name="x density", yaxis="y2")
        else:
            bin_size = density_dict["names"][1] - density_dict["names"][0]
            trace1 = go.Histogram(
                x=data_dict["x"],
                name="x density",
                yaxis="y2",
                autobinx=False,
                xbins=dict(
                    start=density_dict["names"][0],
                    end=density_dict["names"][-1],
                    size=bin_size,
                ),
            )

and then move line 154:
            resp_bin_size = density_dict["names"][1] - density_dict["names"][0]

create a new if statement, and put this line there: (line 157)
        if is_categorical:
            trace2 = go.Histogram(x=data_dict["y"], name="y density", xaxis="x2")
        else:
            resp_bin_size = density_dict["names"][1] - density_dict["names"][0]
            trace2 = go.Histogram(
                y=data_dict["y"],
                name="y density",
                xaxis="x2",
                autobiny=False,
                ybins=dict(
                    start=resp_density_dict["names"][0],
                    end=resp_density_dict["names"][-1],
                    size=resp_bin_size,
                ),
            )

this would resolve the issue.

jbbqqf:
## Summary

`Marginal(...).explain_data(...).visualize(key)` crashed with
`TypeError: unsupported operand type(s) for -: 'str' and 'str'` whenever
`feature_types[key]` was `"nominal"` or `"ordinal"`, making the marginal
data explainer unusable on every dataset that contained a categorical
feature.

Fixes interpretml/interpret#119 — *visualize has a bug for categorical features*

## Context

`MarginalExplanation.visualize` builds a small two-panel plotly figure
(feature histogram + response histogram). It computed the histogram
bin width for both panels with a single line at the top of the function:

```python
bin_size = density_dict["names"][1] - density_dict["names"][0]
```

For continuous features `density_dict["names"]` is the numeric output
of `np.histogram(...)` (bin edges), so the subtraction is fine. For
nominal/ordinal features the same field stores string category labels
returned by `np.unique(...)`, and the subtraction blows up before any
branch on `is_categorical` ever executes — so neither path renders.

There is a second instance of the same pattern further down for the
response axis (`resp_bin_size = density_dict["names"][...]` — note that
it even pointed at the *feature* density rather than the *response*
density). Both are addressed in this PR.

The fix matches the long-standing report from issue #119, with one
small addition: classification targets (categorical response) now also
render, because the response side is now guarded the same way.

## Changes

- `python/interpret-core/interpret/data/_response.py`
  - move `bin_size` into the `else` branch so it is only computed when
    `density_dict["names"]` are numeric bin edges.
  - mirror the same guard on the response axis. When the response is
    categorical (e.g. classification target) build a plotly `Histogram`
    without `ybins`, otherwise compute `resp_bin_size` from
    `resp_density_dict["names"]` (was previously computed off the
    *feature* density — a latent bug the same line fixed).
  - add a tiny `_is_numeric_bin_edges` predicate so the categorical
    response check stays self-documenting; comment cites issue #119
    so future readers don't have to re-derive the reasoning.
- `python/interpret-core/tests/data/test_marginal_visualize.py` (new)
  - four pytest cases: nominal feature on continuous target, continuous
    feature (non-regression), nominal feature on classification target,
    and the `key=None` overall path.
  - the two nominal-feature cases fail on `origin/main` with the exact
    `TypeError` cited in #119; all four pass on this branch.

## Reproduce BEFORE/AFTER yourself (copy-paste)

```bash
# --- one-time setup ---
git clone https://github.com/interpretml/interpret.git /tmp/repro && cd /tmp/repro
python -m venv .venv && source .venv/bin/activate
pip install -e "python/interpret-core[testing]" plotly

cat > /tmp/repro_119.py <<'PY'
import numpy as np
import pandas as pd
from interpret.data import Marginal

df = pd.DataFrame({
    "cat":  ["a", "b", "a", "c", "b", "a", "c", "b"],
    "cont": [1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0],
})
y = np.array([0.5, 1.5, 2.5, 3.5, 4.5, 5.5, 6.5, 7.5])
m = Marginal(feature_types=["nominal", "continuous"]).explain_data(df, y)
print("categorical viz:", type(m.visualize(0)).__name__)
print("continuous viz: ", type(m.visualize(1)).__name__)
PY

# --- BEFORE (origin/main) ---
git checkout origin/main
python /tmp/repro_119.py
# Expected: TypeError: unsupported operand type(s) for -: 'str' and 'str'

# --- AFTER (this PR) ---
git fetch https://github.com/jbbqqf/interpret.git feat/119-fix-marginal-categorical-visualize
git checkout FETCH_HEAD
python /tmp/repro_119.py
# Expected: both lines print "Figure"

# Optional: run the new regression suite, which fails on main and
# passes on this branch
git checkout origin/main
python -m pytest -vv python/interpret-core/tests/data/test_marginal_visualize.py
# Expected: 2 failed, 2 passed (the categorical cases hit the bug)
git checkout FETCH_HEAD
python -m pytest -vv python/interpret-core/tests/data/test_marginal_visualize.py
# Expected: 4 passed
```

## What I ran locally

- `python -m pytest -vv tests/data/test_marginal_visualize.py` → **4 passed**
  on this branch.
- The same suite on `origin/main` → **2 failed, 2 passed** (the two
  failures are the exact `TypeError` cited in #119; the passing ones
  cover the continuous and overall paths that were already correct).
- `ruff check interpret-core/interpret/data/_response.py` → 6 errors
  reported, identical to `origin/main` (all pre-existing
  `PLC0415`/`NPY002` flags in this file). My change introduces zero
  new lint findings.
- `ruff format --check` on the touched files → already formatted.

The selenium / dashboard / full visual suites need extra optional
dependencies (`requests`, `dash`, etc.) that aren't part of the
`testing` extra; they were not run.

## Edge cases tested

| # | Scenario | Input | Expected | Verified by |
|---|----------|-------|----------|-------------|
| 1 | Nominal feature, continuous target | `feature_types=["nominal","continuous"]`, key=0 | plotly Figure (was TypeError) | `test_visualize_nominal_feature_continuous_target_succeeds` |
| 2 | Continuous feature (regression guard) | same dataset, key=1 | plotly Figure (still works) | `test_visualize_continuous_feature_still_works` |
| 3 | Nominal feature, classification target | y = `[0,1,0,1,0,1,0,1]`, key=0 | plotly Figure (categorical response branch) | `test_visualize_nominal_feature_classification_target` |
| 4 | Overall view (`key=None`) | any | plot_density figure (path unchanged) | `test_visualize_overall_unaffected` |

## Risk / blast radius

The change is local to `MarginalExplanation.visualize`. It only adds
code paths that were previously unreachable (categorical features /
classification responses). Continuous-feature behaviour is byte-equal
to before — same plotly trace, same `xbins` dict. The new
`_is_numeric_bin_edges` helper is module-private. No public API
changes.

## Release note

```release-note
Fixed Marginal data explainer crashing with TypeError when visualising
nominal/ordinal features or classification targets (issue #119).
```

---

*PR drafted with assistance from Claude Code. The change was reviewed
manually against `interpretml/interpret`'s source and the cited issue.
The reproducer block above was used during development; it is the same
one a reviewer can paste verbatim.*

### DIFF
--- a/python/interpret-core/interpret/data/_response.py
+++ b/python/interpret-core/interpret/data/_response.py
@@ -10,6 +10,21 @@ from ..utils._explanation import gen_global_selector, gen_name_from_class
 from ..utils._unify_data import unify_data
 
 
+def _is_numeric_bin_edges(names):
+    """Return True iff `names` looks like a numeric bin-edge sequence.
+
+    For continuous features the density payload stores numeric bin edges
+    (e.g. ``[0.0, 1.5, 3.0, ...]``); for categorical features it stores the
+    string category labels themselves (e.g. ``["a", "b", "c"]``). The
+    histogram path computes ``names[1] - names[0]`` to derive the plotly
+    bin size, which raises ``TypeError`` on the categorical case (Issue #119).
+    Use this guard before doing the subtraction.
+    """
+    return len(names) >= 2 and all(
+        isinstance(n, (int, float, np.integer, np.floating)) for n in names[:2]
+    )
+
+
 class Marginal(DataExplainer):
     """Provides a marginal plot for provided data."""
 
@@ -200,15 +215,18 @@ class MarginalExplanation(BaseExplanation):
 
         # Show feature graph
         density_dict = data_dict["feature_density"]
-
-        bin_size = density_dict["names"][1] - density_dict["names"][0]
         is_categorical = (
             self.feature_types[key] == "nominal" or self.feature_types[key] == "ordinal"
         )
 
+        # Bin sizing only makes sense for numeric histograms. For categorical
+        # features density_dict["names"] holds string category labels (Issue #119),
+        # so subtracting them raises TypeError. Compute bin_size lazily only on
+        # the numeric branch.
         if is_categorical:
             trace1 = go.Histogram(x=data_dict["x"], name="x density", yaxis="y2")
         else:
+            bin_size = density_dict["names"][1] - density_dict["names"][0]
             trace1 = go.Histogram(
                 x=data_dict["x"],
                 name="x density",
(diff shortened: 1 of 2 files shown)


paulbkoch:
Very nice PR.  Thanks @jbbqqf!

Issue #626, opened 2025-08-07
Smoothing only for some features defined by the user


m-maggi:
Hi, thanks for the great package.
I would like to use the `smoothing_rounds` argument only for a subset of the features of my model, the rest of the variables should have `smooting_rounds`=0. This since I know which features are smooth and which features I want to model "un-smoothly".
I looked at the examples here https://interpret.ml/docs/python/examples/interpretable-regression-synthetic.html and here https://interpret.ml/docs/python/examples/custom-interactions.html. 
Currently I am merging two models, one where I exclude all the non-smooth features, and one where I exclude all the smooth features (I do this using the argument `exclude`).
I follow the steps of the "custom interactions" example, adjusting the bin weights / scaling / adding intercepts.
I don't know the algorithms in detail: would the described model merge be enough or would one rather need an iterative procedure where mod_smooth and mod_non_smooth are trained every time with 1 boosting round and they are fitted using `init_score` using the values from the previous iterations? With such a procedure both the smooth and the non-smooth model would be always updated rather than only one model being fully trained and then the second model being trained using `init_score`.

paulbkoch:
Hi @m-maggi -- The methodology I proposed above should be either identical or nearly identical to how it would work if incorporated as a new parameter.  Fitting two separate EBMs with different subsets of the features won't be quite as performant, especially when there is correlation between the smooth and non-smooth features.  Also, in that case both of the EBMs are fully fit, so you wouldn't adjust the scores or weights before merging.

Since there is a way to achieve this externally, I'd be hesitant to add a new parameter to the already long list of existing parameters.  I could see modifying the smoothing_rounds parameter to accept a list of integers where the list of integers is the number of smoothing rounds for each feature.  This wouldn't introduce a new parameter.

Eg: smoothing_rounds = [1000, 1000, 0, 0, 0, 1000]  # smooth features 0, 1, and 5.

It's a bit trickier if the user specifies different numbers of smoothing rounds for each feature, but I guess in that case the right thing to do would be to continue to do smoothing while any feature hasn't reached its specified number of smoothing rounds.

paulbkoch:
The issue with having too many bins in the histograms can be fixed.  merge_ebms doesn't yet merge the histograms, which are only used in the visualizations and are shown at the bottom of the graphs.  I added two lines to my code above to eliminate the histograms on the merged model, which makes it more visualizable.

In terms of the model that gets generated, I'm not seeing the same issues in the shape functions.  Here's my feature 4:

[...]

PR #672, opened 2026-05-26
Add per-feature smoothing_rounds support (#626)

ugbotueferhire:
Extend smoothing_rounds and interaction_smoothing_rounds to accept a sequence of ints, one per term, in addition to the existing scalar form. Per Paul's spec on the issue thread: the smoothing phase as a whole runs while any term still has rounds remaining; a term whose counter has reached zero falls back to normal gain-based updates for the remaining smoothing iterations.

* _boost.py: replace the scalar smoothing_rounds counter with a per-term np.int64 array. Scalars are broadcast at the top of the loop, so the inner control flow is uniform. The RandomSplits flag, the make_progress predicate, the early-stopping skip, and the end-of-cycle decrement all become per-term checks.
* _ebm.py: validate int | Sequence[int] in _normalize_smoothing_rounds. At the mains call site, gather the per-feature list down to the post-exclude term_features order. At the interactions call site, only accept a list when interactions is an explicit list (FAST-discovered interactions cannot be aligned to a user-supplied per-interaction parameter), and align via boost_groups_user_idx to handle dedup.
* Update docstrings on ExplainableBoostingClassifier/Regressor and document that DP-EBM ignores the list form.

Tests cover scalar/list equivalence, per-feature zero disables smoothing for those features without breaking other features' smoothing, exclude alignment, validation errors (wrong length, negatives, non-integers, empty), and the interactions-list constraint.

### DIFF
--- a/python/interpret-core/interpret/glassbox/_ebm.py
+++ b/python/interpret-core/interpret/glassbox/_ebm.py
@@ -221,6 +221,73 @@ def _normalize_callbacks(callbacks):
     return found.get("examine"), found.get("boost"), found.get("interaction")
 
 
+def _normalize_smoothing_rounds(
+    value: float | Sequence[int] | npt.NDArray[Any] | None,
+    name: str,
+) -> int | npt.NDArray[np.int64]:
+    """Normalize a smoothing_rounds parameter.
+
+    Returns an ``int`` when the caller passed a scalar (or ``None``, which
+    is treated as zero); returns a 1-D int64 ``ndarray`` when the caller
+    passed a sequence. Keeping these distinct lets call sites enforce that
+    a sequence's length match the term count (a length-1 sequence is NOT
+    silently broadcast). Raises ``ValueError`` on negatives, non-integers,
+    or wrong dimensionality.
+    """
+    if value is None:
+        return 0
+
+    if isinstance(value, (int, np.integer)) and not isinstance(value, bool):
+        scalar = int(value)
+        if scalar < 0:
+            msg = f"{name} cannot be negative"
+            _log.error(msg)
+            raise ValueError(msg)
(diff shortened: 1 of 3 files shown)


ugbotueferhire:
hello @paulbkoch how are you doing, 
I made submitted a PR will be awaiting your feedback :)

ugbotueferhire:
Hello @paulbkoch still waiting on your feedback

python/interpret-core/interpret/glassbox/_ebm_core/_boost.py
```diff
@@ -181,7 +194,7 @@ def boost(
                             term_idx = random_cyclic_ordering[state_idx]
 
                             make_progress = False
-                            if cyclic_state >= 1.0 or smoothing_rounds > 0:
+                            if cyclic_state >= 1.0 or smoothing_counts[term_idx] > 0:
                                 # if cyclic_state is above 1.0 we make progress
                                 make_progress = True
```
paulbkoch, line 220: I think this works, but is computationally wasteful.  When make_progress is True we compute the update but then do not apply it.  Instead of doing that, we should advance state_idx to progress to the next term.

python/interpret-core/interpret/glassbox/_ebm.py
paulbkoch: also need to update these to allow sequences

paulbkoch (changes requested):

ugbotueferhire:
hello @paulbkoch I made an update based on your feedback please review when you are free :)

ugbotueferhire:
hello @paulbkoch still waiting on your feedback

ugbotueferhire:
hello @paulbkoch

